PR #17, opened 2022-11-22
Pull Request

seanygberg:
This is my implementation for some of the operators. I still have to figure out how to put these into the interpreter, but this is my current syntax.

### DIFF
--- a/src/Adder/Lang/Lexer.hs
+++ b/src/Adder/Lang/Lexer.hs
@@ -40,7 +40,7 @@ adderLexSpec =
       Tok.opStart = Tok.opLetter adderLexSpec,
       Tok.opLetter = oneOf "+-*/%@<>&|^~:=!.",
       Tok.reservedOpNames =
-        ["%", "//"]
+        ["%", "//", "/"]
     }
 
 boolean :: IParser Bool
--- a/src/Adder/Lang/Lexer.hs
+++ b/src/Adder/Lang/Lexer.hs
@@ -40,7 +40,7 @@ adderLexSpec =
       Tok.opStart = Tok.opLetter adderLexSpec,
       Tok.opLetter = oneOf "+-*/%@<>&|^~:=!.",
       Tok.reservedOpNames =
-        []
+        ["%", "//"]
     }
 
 boolean :: IParser Bool
--- a/src/Adder/Lang/Lexer.hs
+++ b/src/Adder/Lang/Lexer.hs
@@ -40,7 +40,7 @@ adderLexSpec =
       Tok.opStart = Tok.opLetter adderLexSpec,
       Tok.opLetter = oneOf "+-*/%@<>&|^~:=!.",
       Tok.reservedOpNames =
-        ["%", "//", "/"]
+        ["*","%", "//", "/"]
     }
 
 boolean :: IParser Bool
--- a/src/Adder/Lang/Parser.hs
+++ b/src/Adder/Lang/Parser.hs
@@ -71,7 +71,6 @@ simpleStmt =
 stmtList :: IParser Statement
 stmtList = undefined
 
-
 -- Implementation of modulo
 -- EBNF Rule: operator ::== "%"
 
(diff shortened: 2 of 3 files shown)


Trent325 (commented):
The addition of the operators and their parser looks correct. I am not entirely sure but I believe it is.

rvarca22:
Would you not also have to add the operators '//' and '%' into the operator precedence table in Parser? I believe mod takes precedence over + and -

Gjidodaj02:
From what I see it looks like you implemented them well. Not too sure though since I settled with implementing easier operators lol, so its harder for me to compare.

src/Adder/Lang/Lexer.hs
```diff
@@ -40,7 +40,7 @@ adderLexSpec =
       Tok.opStart = Tok.opLetter adderLexSpec,
       Tok.opLetter = oneOf "+-*/%@<>&|^~:=!.",
       Tok.reservedOpNames =
-        []
+        ["%", "//", "/"]
```
Tientuine, line 43: FYI, someone else (Tevie) is also working on remainder/modulus because you never "claimed" it in the Issues (i.e., wrote your name next to it).

Otherwise, this looks fine.

src/Adder/Lang/Parser.hs
```diff
@@ -71,13 +71,29 @@ simpleStmt =
...
   [ [Prefix (reservedOp "-" >> return (UnaryExpr Negative))],
     [Infix (reservedOp "**" >> return (BinaryExpr Power)) AssocRight],
     [ Infix (reservedOp "*" >> return (BinaryExpr Times)) AssocLeft
     ],
+    [ Infix (reservedOp "/" >> return (BinaryExpr Divide)) AssocLeft
+    ],
+    [ Infix (reservedOp "//" >> return (BinaryExpr IntDiv)) AssocLeft
+    ],
+    [ Infix (reservedOp "%" >> return (BinaryExpr Mod)) AssocLeft
+    ],
```
Tientuine, line 96: To make sure that these operators all have the same precedence, which they should, you must put them together in the same "row" of the `table`. In other words, they should all be in the same sub-list, rather than in separate lists.

src/Adder/Lang/Syntax.hs
```diff
@@ -39,6 +39,9 @@ data BinaryOp
   | Less
   | And
   | Is
+  | Mod
+  | IntDiv
+  | Divide
```
Tientuine, line 44: Good.

Tientuine (changes requested):
Almost there!

src/Adder/Lang/Parser.hs
```diff
@@ -86,14 +86,10 @@ table :: [[Operator String () (IndentT Identity) Expression]]
...
-    [ Infix (reservedOp "/" >> return (BinaryExpr Divide)) AssocLeft
-    ],
-    [ Infix (reservedOp "//" >> return (BinaryExpr IntDiv)) AssocLeft
-    ],
-    [ Infix (reservedOp "%" >> return (BinaryExpr Mod)) AssocLeft
-    ],
+    [ Infix (reservedOp "*" >> return (BinaryExpr Times)) AssocLeft,
+    Infix (reservedOp "/" >> return (BinaryExpr Divide)) AssocLeft,
+    Infix (reservedOp "//" >> return (BinaryExpr IntDiv)) AssocLeft,
+    Infix (reservedOp "%" >> return (BinaryExpr Mod)) AssocLeft],
```
Tientuine, line 92: Better!

src/Adder/Lang/Syntax.hs
```diff
@@ -42,7 +42,7 @@ data BinaryOp
   | Mod
   | IntDiv
   | Divide
-  deriving (Show)
+  deriving (Eq, Ord, Show)
```
Tientuine, line 45: Not really needed, but fine.

src/Adder/Lang/Lexer.hs
Tientuine, line 43: Actually, we also need the multiplication operator `*` in this list.

Tientuine (changes requested):
Just add the `*` operator to the list of reserved operator names, and then you're all set for merge!

Tientuine (approved):
Yay!

PR #18, opened 2022-11-22
Assignment 3

Trent325:
I attempted to do assignment 3, it is probably wrong, so feedback is appreciated. Thank you

### DIFF
--- a/src/Adder/Lang/Parser.hs
+++ b/src/Adder/Lang/Parser.hs
@@ -45,11 +45,11 @@ contents p = do
 
 -- See https://docs.python.org/3/reference/toplevel_components.html#complete-python-programs
 program :: IParser Program
-program = undefined
+program = Pgm <*> statement
 
 -- See https://docs.python.org/3/reference/compound_stmts.html#grammar-token-python-grammar-statement
 statement :: IParser Statement
-statement = undefined
+statement = [compoundStmt, simpleStmt]
 
 -- See https://docs.python.org/3/reference/compound_stmts.html#grammar-token-python-grammar-suite
 suite :: IParser [Statement]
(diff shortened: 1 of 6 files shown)


seanygberg:
I think you have the right idea as to how everything should be implemented. The one thing I will say is when you define it in simple statement in the Parser, it think you just need (reserved "return" >> [expression] ) instead of (reserved "return" >> return [expression] ). I could be wrong, but I think that is how you could do it.

rvarca22:
Looks good to me. You should probably name your expression in Syntax "ReturnExpr" just to keep the naming convention for expressions in the language. Not sure if you're allowed to have a space in the name of an expression anyway. 

I also agree with Sean's comment, the return statement in the Parser should read "reserved 'return' >> Expression" you don't need the second return

Tientuine (changes requested):

src/Adder/Lang/Parser.hs
```diff
@@ -45,11 +45,16 @@ contents p = do
...
 
 -- See https://docs.python.org/3/reference/compound_stmts.html#grammar-token-python-grammar-statement
 statement :: IParser Statement
-statement = undefined
+statement = 
+  (choice . map try)
+    [
+    compoundStmt --Attempted to Fix After Johnson feedback
+    
+    ]
```
Tientuine, line 57: Fine. Formatting is off, but at this point I'll just fix it when resolving the merge conflicts.

src/Adder/Lang/Parser.hs
```diff
@@ -61,11 +66,16 @@ compoundStmt =
   (choice . map try)
     []
 
+
+
```
Tientuine, line 70: Don't really need these extra empty lines, but okay.

src/Adder/Lang/Parser.hs
```diff
@@ -61,11 +66,16 @@ compoundStmt =
...
 -- See https://docs.python.org/3/reference/simple_stmts.html#grammar-token-python-grammar-simple_stmt
 simpleStmt :: IParser Statement
 simpleStmt =
   (choice . map try)
-    []
+    [
+      ReturnStmt <$> (reserved "return" >> expression)  --  (reserved "return" >> [Expression]) -- Attempted to make it like the IsZero expression after feedback 
+      -- Attempted EBNF rule return_stmt ::=  "return" [expression_list]
+    ]
```
Tientuine, line 78: Good for now.

src/Adder/Lang/Syntax.hs
```diff
@@ -16,7 +16,7 @@ newtype Program
 
 -- TODO Build out the abstract syntax for Adder statements
 data Statement
-  = NoStmt -- TODO Remove this after adding legit data constructors
+  =  ReturnStmt Expression  -- I did this to return just an expression of any kind, so the user could use return for many purposes -- TODO Remove this after adding legit data constructors
```
Tientuine, line 19: Looks good for now.

Tientuine (approved):
A couple of formatting issues, but I'll just fix these when doing the merge. In the meantime, try to figure out the VSCode + Haskell extensions problem so all future work can be formatting properly.

PR #19, opened 2022-11-22
Assignment 3 - Marsh

IanMarsh1:
Still have issues with the Parser.hs file and trying to get the stack. Error below:

> src\Adder\Lang\Parser.hs:48:19: error:

    * Couldn't match type Statement with [Statement]
      Expected: ParsecT String () (IndentT Identity) [Statement]
        Actual: IParser Statement
    * In the second argument of (<$>), namely statement
      In the expression: Pgm <$> statement
      In an equation for `program': program = Pgm <$> statement
     |
48 | program = Pgm <$> statement
     |______________________^^^^^^^^^

### DIFF
--- a/src/Adder/Lang/Lexer.hs
+++ b/src/Adder/Lang/Lexer.hs
@@ -34,7 +34,7 @@ adderLexSpec =
       -- TODO Define the reserved names/keywords for the Adder language
       -- See https://docs.python.org/3/reference/lexical_analysis.html#keywords
       Tok.reservedNames =
-        [],
+        ["pass"],
       -- TODO Define the reserved operator symbols for the Adder language
       -- See https://docs.python.org/3/reference/lexical_analysis.html#operators
       Tok.opStart = Tok.opLetter adderLexSpec,
--- a/src/Adder/Lang/Lexer.hs
+++ b/src/Adder/Lang/Lexer.hs
@@ -34,13 +34,13 @@ adderLexSpec =
       -- TODO Define the reserved names/keywords for the Adder language
       -- See https://docs.python.org/3/reference/lexical_analysis.html#keywords
       Tok.reservedNames =
-        ["pass"],
+        ["pass", "return", "if", "or", "and", "is", "not", "in", "not in", "is not"],
       -- TODO Define the reserved operator symbols for the Adder language
       -- See https://docs.python.org/3/reference/lexical_analysis.html#operators
       Tok.opStart = Tok.opLetter adderLexSpec,
       Tok.opLetter = oneOf "+-*/%@<>&|^~:=!.",
-      Tok.reservedOpNames =
-        []
+      Tok.reservedOpNames = 
+        ["**", "-", "*", "%", "//", "/", "+", "<", "<=", ">", ">=", "!=", "=="]
     }
 
 boolean :: IParser Bool
--- a/src/Adder/Lang/Parser.hs
+++ b/src/Adder/Lang/Parser.hs
@@ -18,8 +18,7 @@ import Adder.Lang.Lexer
 import Adder.Lang.Syntax
 import Control.Monad (liftM2)
 import Data.Functor.Identity (Identity)
-import Text.Parsec hiding (parse, string, sepBy)
---import Text.Parsec (ParseError, choice, eof, many, sepBy, try)
+import Text.Parsec hiding (parse, string)
 import Text.Parsec.Expr
 import Text.Parsec.Indent
 
@@ -46,11 +45,7 @@ contents p = do
 
 -- See https://docs.python.org/3/reference/toplevel_components.html#complete-python-programs
 program :: IParser Program
-program = Pgm <$> statement --help
-
---block :: IParser a -> IParser [a]  ?????
---block identifier :: IParser [String]
---block integer :: IParser [Integer]
+program = Pgm <$> block statement
 
 -- See https://docs.python.org/3/reference/compound_stmts.html#grammar-token-python-grammar-statement
 statement :: IParser Statement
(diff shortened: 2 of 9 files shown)


Tientuine:
> Still have issues with the Parser.hs file and trying to get the stack. Error below:
> 
> > src\Adder\Lang\Parser.hs:48:19: error:
> 
> ```
> * Couldn't match type Statement with [Statement]
>   Expected: ParsecT String () (IndentT Identity) [Statement]
>     Actual: IParser Statement
> * In the second argument of (<$>), namely statement
>   In the expression: Pgm <$> statement
>   In an equation for `program': program = Pgm <$> statement
>  |
> ```
> 
> 48 | program = Pgm <$> statement |______________________^^^^^^^^^

It's useful to note here that `ParsecT String () (IndentT Identity)` means the same thing as `IParser` in this context. So, what it's telling you is that `Pgm <$> statement` gives you an `IParser Statement`, but what it needs is an `IParser [Statement]`. That's because the `statement` function is an `IParser Statement`.

This is where the `block` function that I mentioned earlier in the Zoom would be useful. The type of the block function is `block :: IParser a -> IParser [a]`. In other words, if you give is a function that parses an `a` value, then it creates a function that parses an sequence of `a`s on separate lines, as long as they're all indented the same amount (which is what we want).

seanygberg:
The pass statement looks to be implemented correctly. I don't know if you need to add "return PassStmt" when you define it in the parser. I could be wrong, but I don't think you need it.

src/Adder/Lang/Lexer.hs
```diff
@@ -34,7 +34,7 @@ adderLexSpec =
       -- TODO Define the reserved names/keywords for the Adder language
       -- See https://docs.python.org/3/reference/lexical_analysis.html#keywords
       Tok.reservedNames =
-        [],
+        ["pass"],
```
Tientuine, line 37: Good!

src/Adder/Lang/Parser.hs
```diff
@@ -45,11 +45,16 @@ contents p = do
 
 -- See https://docs.python.org/3/reference/toplevel_components.html#complete-python-programs
 program :: IParser Program
-program = undefined
+program = Pgm <$> statement
```
Tientuine, line 48: It's useful to note here that `ParsecT String () (IndentT Identity)` means the same thing as `IParser` in this context. So, what it's telling you is that `Pgm <$> statement` gives you an `IParser Statement`, but what it needs is an `IParser [Statement]`. That's because the `statement` function is an `IParser Statement`.

This is where the `block` function that I mentioned earlier in the Zoom would be useful. The type of the `block` function is `block :: IParser a -> IParser [a]`. In other words, if you give it a function that parses an `a` value, then it creates a function that parses an sequence of `a`s on separate lines, as long as they're all indented the same amount (which is what we want).

As an example, if we write `block identifier` that will have type `IParser [String]` and will parse an indented block of `identifier`s, each identifier on its own line and all lines indented the same amount. Of course, we don't want a block of identifiers... we want a block of `statement`s.

src/Adder/Lang/Parser.hs
```diff
@@ -45,11 +45,16 @@ contents p = do
...
 
 -- See https://docs.python.org/3/reference/compound_stmts.html#grammar-token-python-grammar-statement
 statement :: IParser Statement
-statement = undefined
+statement = 
+  (choice . map try)
+    [
+      compoundStmt, 
+      simpleStmt
+    ]
```
Tientuine, line 57: This is very close... however, if we look at the Python language documentation, a `statement` can be either a `compoundStmt` or a `stmtList` (not just a single `simpleStmt`).

Tientuine (changes requested):
A few things need fixing (see below), but this is close and a good start.

src/Adder/Lang/Parser.hs
```diff
@@ -45,11 +45,15 @@ contents p = do
 
 -- See https://docs.python.org/3/reference/toplevel_components.html#complete-python-programs
 program :: IParser Program
-program = undefined
+program = Pgm <$> block statement
```
Tientuine, line 48: This change already in other branches, but okay. In the future, we should try to make sure that only one person makes a given change, so smoother merging.

src/Adder/Lang/Parser.hs
```diff
@@ -45,11 +45,15 @@ contents p = do
...
 
 -- See https://docs.python.org/3/reference/compound_stmts.html#grammar-token-python-grammar-statement
 statement :: IParser Statement
-statement = undefined
+statement =
+  (choice . map try)
+    [ compoundStmt,
+      stmtList
+    ]
```
Tientuine, line 56: Looks good.

src/Adder/Lang/Parser.hs
```diff
@@ -65,11 +69,16 @@ compoundStmt =
 simpleStmt :: IParser Statement
 simpleStmt =
   (choice . map try)
-    []
+    [ (reserved "pass" >> return PassStmt) -- pass_stmt ::= "pass"
```
Tientuine, line 72: Correct, but the parentheses are not needed.

src/Adder/Lang/Parser.hs
```diff
@@ -65,11 +69,16 @@ compoundStmt =
...
+    ]
 
 -- See https://docs.python.org/3/reference/compound_stmts.html#grammar-token-python-grammar-stmt_list
 stmtList :: IParser Statement
-stmtList = undefined
+stmtList =
+  (choice . map try)
+    [ StmtList
+        <$> (sepBy simpleStmt (symbol ";"))
+    ]
```
Tientuine, line 81: Good. Technically not quite correct, since your code allows a statement list to be zero statements, which is not permitted in Python, but we can fix that later. Again, the outer parentheses are not necessary.

src/Adder/Lang/Syntax.hs
```diff
@@ -16,7 +16,8 @@ newtype Program
 
 -- TODO Build out the abstract syntax for Adder statements
 data Statement
-  = NoStmt -- TODO Remove this after adding legit data constructors
+  = PassStmt
+  | StmtList [Statement]
```
Tientuine, line 20: Good.

Tientuine (approved):
Looks ready for merge.

PR #20, opened 2022-11-22
Varca work

rvarca22:


Gjidodaj02:
From what I quickly scanned through, your implementation looks good.

src/Adder/Interp.hs
```diff
@@ -50,6 +50,10 @@ resultOfProgram _ env st0 = undefined
 
 resultOf :: Statement -> Environment -> Store -> IO Store
 resultOf _ env st0 = undefined
+resultOf(IfStmt test conseq) p st = if q then st2 --eventually 'else st3'
+  where 
+    Answer (BoolVal q) st1 = valueOf test p st
+    st2 = resultOf conseq p st1
```
Tientuine, line 56: Nearly there... Haskell if-expressions need an else no matter what, you can't leave it out. So, for an if-statement that doesn't have an else block, what will be the resulting state if the test if false?

Also, an if-statement in Python will have a `suite` following the test, and a `suite` is a basically list of statements. So, the `conseq` symbol will actually refer to a list of statements rather than a single statement. Consequently (pun intended), you'll need to compute the result of each of these in sequence and then return the final state.

src/Adder/Lang/Parser.hs
```diff
@@ -45,11 +45,11 @@ contents p = do
 
 -- See https://docs.python.org/3/reference/toplevel_components.html#complete-python-programs
 program :: IParser Program
-program = undefined
+program = Pgm <$> statement
```
Tientuine, line 48: Try to avoid making the same changes that another student is making (Ian also has this change in his branch), otherwise we will have merge conflicts.

Here are my comments to him about this one:

It's useful to note here that `ParsecT String () (IndentT Identity)` means the same thing as `IParser` in this context. So, what it's telling you is that `Pgm <$> statement` gives you an `IParser Statement`, but what it needs is an `IParser [Statement]`. That's because the `statement` function is an `IParser Statement`.

This is where the `block` function that I mentioned earlier in the Zoom would be useful. The type of the `block` function is `block :: IParser a -> IParser [a]`. In other words, if you give it a function that parses an `a` value, then it creates a function that parses an sequence of `a`s on separate lines, as long as they're all indented the same amount (which is what we want).

As an example, if we write `block identifier` that will have type `IParser [String]` and will parse an indented block of `identifier`s, each identifier on its own line and all lines indented the same amount. Of course, we don't want a block of identifiers... we want a block of `statement`s.

src/Adder/Lang/Syntax.hs
```diff
@@ -16,7 +16,8 @@ newtype Program
 
 -- TODO Build out the abstract syntax for Adder statements
 data Statement
-  = NoStmt -- TODO Remove this after adding legit data constructors
+  = IfStmt  expression [statement]   -- If Statement contains Exprression and list of statements - at this time only one statement
```
Tientuine, line 19: Right idea, but you need capitalized `Expression` and `[Statement]` here because we need to refer to the types (not the functions in the Parser.hs file).

Tientuine (changes requested):
Good attempt here, very close - there are just a few fixes needed, see below for explanation and suggestions.

src/Adder/Lang/Syntax.hs
rvarca22: Resolved

src/Adder/Lang/Parser.hs
rvarca22: resolved

src/Adder/Lang/Lexer.hs
```diff
@@ -34,11 +34,11 @@ adderLexSpec =
       -- TODO Define the reserved names/keywords for the Adder language
       -- See https://docs.python.org/3/reference/lexical_analysis.html#keywords
       Tok.reservedNames =
-        [],
+        ["if"],
       -- TODO Define the reserved operator symbols for the Adder language
       -- See https://docs.python.org/3/reference/lexical_analysis.html#operators
       Tok.opStart = Tok.opLetter adderLexSpec,
-      Tok.opLetter = oneOf "+-*/%@<>&|^~:=!.",
+      Tok.opLetter = oneOf "+-*/%@<>&|^~ := !.",
```
Tientuine, line 41: This is incorrect, there should be no whitespace among the operator letters.

src/Adder/Lang/Parser.hs
```diff
@@ -44,12 +44,15 @@ contents p = do
 -- See https://docs.python.org/3/reference/grammar.html
 
 -- See https://docs.python.org/3/reference/toplevel_components.html#complete-python-programs
+
+--block :: IParser statement -> IParser [statement]
+
 program :: IParser Program
-program = undefined
+program = Pgm <$> block statement
```
Tientuine, line 51: This is correct, though Trent already has this in his branch. In the future, try to avoid having two people making the same changes.

src/Adder/Lang/Parser.hs
```diff
@@ -44,12 +44,15 @@ contents p = do
...
 
 -- See https://docs.python.org/3/reference/compound_stmts.html#grammar-token-python-grammar-statement
 statement :: IParser Statement
-statement = undefined
+statement = compoundStmt -- for my
```
Tientuine, line 55: Okay.

src/Adder/Lang/Parser.hs
```diff
@@ -94,6 +101,13 @@ table =
 expression :: IParser Expression
 expression = buildExpressionParser table atom <?> "expression"
 
+--assignment_expression ::=  [identifier ":="] expression
+assignmentExpr :: IParser Expression
+assignmentExpr = expression --For now, assignment expression only needs to be an expression
+--(choice . map try)
+--  [<$> identifier
+--    <*> (reservedOp "=" >> Expression)]
+
```
Tientuine, line 110: Looks good.

src/Adder/Lang/Syntax.hs
```diff
@@ -16,7 +16,8 @@ newtype Program
 
 -- TODO Build out the abstract syntax for Adder statements
 data Statement
-  = NoStmt -- TODO Remove this after adding legit data constructors
+  = IfStmt Expression [Statement] -- If Statement contains Exprression and list of statements - at this time only one statement
```
Tientuine, line 19: Good for now. Later we'll make this if-statement more flexible to handle other branches.

Tientuine (approved):
Technically, there is a small mistake in the opLetters list, but I'll just fix that when fixing other merge conflicts so that we can move things along more quickly.

PR #23, opened 2022-11-24
Break Statement

macc125:
Implementation of break statement, with modifications to the following:

- Added reserved Op Names to Lexer
- Added: Negate, Greater, LessEq, GreaterEq, NotEqual operators

### DIFF
--- a/src/Adder/Lang/Lexer.hs
+++ b/src/Adder/Lang/Lexer.hs
@@ -40,7 +40,7 @@ adderLexSpec =
       Tok.opStart = Tok.opLetter adderLexSpec,
       Tok.opLetter = oneOf "+-*/%@<>&|^~:=!.",
       Tok.reservedOpNames =
-        ["Not", "Negative", "IsNeg", "IsPos", "IsNull", "Negate", "Greater", "LessEq", "GreaterEq", "And", "Is", "Negate", "Break"]
+        ["not", "-",  ">", "<=>", ">=", "and", "is"]
     }
 
 boolean :: IParser Bool
--- a/src/Adder/Lang/Lexer.hs
+++ b/src/Adder/Lang/Lexer.hs
@@ -40,7 +40,7 @@ adderLexSpec =
       Tok.opStart = Tok.opLetter adderLexSpec,
       Tok.opLetter = oneOf "+-*/%@<>&|^~:=!.",
       Tok.reservedOpNames =
-        []
+        ["Not", "Negative", "IsNeg", "IsPos", "IsNull", "Negate", "Greater", "LessEq", "GreaterEq", "And", "Is", "Negate", "Break"]
     }
 
 boolean :: IParser Bool
--- a/src/Adder/Lang/Lexer.hs
+++ b/src/Adder/Lang/Lexer.hs
@@ -40,7 +40,7 @@ adderLexSpec =
       Tok.opStart = Tok.opLetter adderLexSpec,
       Tok.opLetter = oneOf "+-*/%@<>&|^~:=!.",
       Tok.reservedOpNames =
-        ["not", "-",  ">", "<=>", ">=", "and", "is"]
+        ["not", "-",  ">", "<=", ">=", "and", "is"]
     }
 
 boolean :: IParser Bool
--- a/src/Adder/Lang/Parser.hs
+++ b/src/Adder/Lang/Parser.hs
@@ -79,42 +79,42 @@ stmtList = undefined
 -- See https://docs.python.org/3/reference/expressions.html#operator-precedence
 table :: [[Operator String () (IndentT Identity) Expression]]
 table =
-[
--- Exponentiation
-[Infix (reservedOp "**" >> return (BinaryExpr Power)) AssocRight],
--- Negative
-[Prefix (reservedOp "-" >> return (UnaryExpr Negative))],
--- Multiplication
-[Infix (reservedOp "*" >> return (BinaryExpr Times)) AssocLeft],
--- addition - string
-[Infix (reservedOp "+" >> return (BinaryExpr Plus)) AssocLeft],
--- Comparative Equals
-[Infix (reservedOp "==" >> return (BinaryExpr Equal)) AssocLeft],
--- Comparative Not Equals
-[Infix (reservedOp "!=" >> return (BinaryExpr NotEqual)) AssocLeft],
+  [
+	  -- Exponentiation
+	[Infix (reservedOp "**" >> return (BinaryExpr Power)) AssocRight],
+	-- Negative
+	[Prefix (reservedOp "-" >> return (UnaryExpr Negative))],
+	-- Multiplication
+	[Infix (reservedOp "*" >> return (BinaryExpr Times)) AssocLeft],
+	-- addition - string
+	[Infix (reservedOp "+" >> return (BinaryExpr Plus)) AssocLeft],
+	-- Comparative Equals
+	[Infix (reservedOp "==" >> return (BinaryExpr Equal)) AssocLeft],
+	-- Comparative Not Equals
+	[Infix (reservedOp "!=" >> return (BinaryExpr NotEqual)) AssocLeft],
 
-[
-    -- less or equal
-    Infix (reservedOp "<=" >> return (BinaryExpr LessEq)) AssocLeft,
-    -- greater or equal
-    Infix (reservedOp ">=" >> return (BinaryExpr GreaterEq)) AssocLeft,
-    -- less
-    Infix (reservedOp "<" >> return (BinaryExpr Less)) AssocLeft,
-    -- greater
-    Infix (reservedOp ">" >> return (BinaryExpr Greater)) AssocLeft,
-    Infix (reserved "is" >> return (BinaryExpr Is)) AssocLeft,
+	[
+	-- less or equal
+	Infix (reservedOp "<=" >> return (BinaryExpr LessEq)) AssocLeft,
+	-- greater or equal
+	Infix (reservedOp ">=" >> return (BinaryExpr GreaterEq)) AssocLeft,
+	-- less
+	Infix (reservedOp "<" >> return (BinaryExpr Less)) AssocLeft,
+	-- greater
+	Infix (reservedOp ">" >> return (BinaryExpr Greater)) AssocLeft,
+	Infix (reserved "is" >> return (BinaryExpr Is)) AssocLeft
 
-],
+	],
 
--- not
-[Prefix (reserved "not" >> return (UnaryExpr Not))],
+	-- not
+	[Prefix (reserved "not" >> return (UnaryExpr Not))],
 
--- and
-[Infix (reserved "and" >> return (BinaryExpr And)) AssocLeft],
+	-- and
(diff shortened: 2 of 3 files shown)


src/Adder/Lang/Lexer.hs
```diff
@@ -40,7 +40,7 @@ adderLexSpec =
       Tok.opStart = Tok.opLetter adderLexSpec,
       Tok.opLetter = oneOf "+-*/%@<>&|^~:=!.",
       Tok.reservedOpNames =
-        []
+        ["Not", "Negative", "IsNeg", "IsPos", "IsNull", "Negate", "Greater", "LessEq", "GreaterEq", "And", "Is", "Negate", "Break"]
```
Tientuine, line 43: These should actually be the symbols themselves, *not* the internal constructor names that we use in the abstract syntax (Syntax.hs).

src/Adder/Lang/Parser.hs
```diff
@@ -74,21 +75,38 @@ stmtList = undefined
...
+[Infix (reservedOp ">=" >> return (BinaryExpr GreaterEq)) AssocLeft],
+-- and
+[Infix (reserved "and" >> return (BinaryExpr And)) AssocLeft].
+-- or
+[Infix (reserved "is" >> return (BinaryExpr Is)) AssocLeft],
+-- not
+[Prefix (reserved "not" >> return (UnaryExpr Not))]
+
+-- negate - integer
+[Prefix (reserved "!" >> return (UnaryExpr Negate))]
```
Tientuine, line 106: Please double check the precedence table using the link in the code comment. the order of some of these operators in our `table` does not match the intended order described in the documentation.

When some operators should have the same precedence, then we can simply place them in the same list, as in...
```haskell
    [ Infix (reserved "is" >> return (BinaryExpr Is)) AssocLeft,
      Infix (reservedOp "<" >> return (BinaryExpr Less)) AssocLeft,
      Infix (reservedOp "==" >> return (BinaryExpr Equal)) AssocLeft
    ],
```

src/Adder/Lang/Parser.hs
```diff
@@ -74,21 +75,38 @@ stmtList = undefined
...
+-- break
+[Prefix (reserved "break" >> return Statement BreakStmt)]
+]
```
Tientuine, line 109: The `break` statement is not an operator, and so it should not be included in this `table`.

Instead, it is a kind of `simpleStmt` and should appear as part of that function.

src/Adder/Lang/Syntax.hs
```diff
@@ -17,17 +17,28 @@ newtype Program
 -- TODO Build out the abstract syntax for Adder statements
 data Statement
   = NoStmt -- TODO Remove this after adding legit data constructors
+  | BreakStmt
```
Tientuine, line 20: Good, but you can now remove the `NoStmt` placeholder now that you are adding a real kind of statement.

src/Adder/Lang/Syntax.hs
```diff
@@ -17,17 +17,28 @@ newtype Program
...
   deriving (Show)
 
 -- TODO Build out the abstract syntax for Adder by adding more expressions
 data Expression
   = UnaryExpr UnaryOp Expression
   | BinaryExpr BinaryOp Expression Expression
+  | BreakExpr ForExpr Expression
```
Tientuine, line 27: This is not necessary.

src/Adder/Lang/Syntax.hs
```diff
@@ -17,17 +17,28 @@ newtype Program
...
+
+
 data UnaryOp
   = Not
   | Negative
+  | IsNeg
+  | IsPos
+  | IsNull
+  | Negate
+  | Break
```
Tientuine, line 39: This is not necessary.

src/Adder/Lang/Syntax.hs
```diff
@@ -17,17 +17,28 @@ newtype Program
...
   deriving (Show)
 
+
+
 data UnaryOp
   = Not
   | Negative
+  | IsNeg
+  | IsPos
+  | IsNull
```
Tientuine, line 37: I don't think that there are operators in Adder (i.e., Python) for these, so they should probably not be included here.

src/Adder/Lang/Syntax.hs
```diff
@@ -36,9 +47,14 @@ data BinaryOp
...
   | Power
   | Equal
+  | NotEqual
   | Less
+  | Greater
+  | LessEq
+  | GreaterEq
   | And
   | Is
+  | Mod
```
Tientuine, line 57: These are good.

Tientuine (changes requested):
There are some fixes needed, but it's good to see that you've started several of these new operators and the break statement. Of course, you won't be able to test the break statement until someone implements a loop.

You (or someone else) will also need to implement the behavior of these operations in the **Interp.hs** file.

src/Adder/Lang/Parser.hs
```diff
@@ -75,38 +79,51 @@ stmtList = undefined
...
+
 -- or
-[Infix (reserved "is" >> return (BinaryExpr Is)) AssocLeft],
--- not
-[Prefix (reserved "not" >> return (UnaryExpr Not))]
+[Infix (reserved "or" >> return (BinaryExpr Or)) AssocLeft],
+
 
 -- negate - integer
 [Prefix (reserved "!" >> return (UnaryExpr Negate))]
```
Tientuine, line 125: I'm not sure what this is supposed to be... Python (and thus Adder) does not have the `!` operator.

Tientuine (changes requested):
Just a couple of very small corrections needed and then this will be ready for merge.

src/Adder/Lang/Parser.hs
macc125: - [ x] Negate removed

src/Adder/Lang/Parser.hs
```diff
@@ -75,38 +79,51 @@ stmtList = undefined
...
-[Prefix (reservedOp "-" >> return (UnaryExpr Negative))],
--- addition string
+[
+[Infix (reservedOp "**" >> return (BinaryExpr Power)) AssocRight],
 [Infix (reservedOp "*" >> return (BinaryExpr Times)) AssocLeft],
+[
+    -- addition - string
+    Infix (reservedOp "+" >> return (BinaryExpr Plus)) AssocLeft,
+    -- subtraction - string
+    Prefix (reservedOp "-" >> return (UnaryExpr Negative))
```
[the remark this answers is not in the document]
macc125: - [ ] Unary Negative moved to proper order of precedence in c9a6928a949f9e78d048b1bdc65a1d7547e457ea

src/Adder/Lang/Parser.hs
```diff
@@ -66,7 +66,11 @@ compoundStmt =
 simpleStmt :: IParser Statement
 simpleStmt =
   (choice . map try)
-    []
+    [
+    BreakStmt -- finish me!
```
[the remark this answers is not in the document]
macc125: Solution attempted in c9a6928a949f9e78d048b1bdc65a1d7547e457ea

src/Adder/Lang/Lexer.hs
```diff
@@ -40,7 +40,7 @@ adderLexSpec =
       Tok.opStart = Tok.opLetter adderLexSpec,
       Tok.opLetter = oneOf "+-*/%@<>&|^~:=!.",
       Tok.reservedOpNames =
-        ["Not", "Negative", "IsNeg", "IsPos", "IsNull", "Negate", "Greater", "LessEq", "GreaterEq", "And", "Is", "Negate", "Break"]
+        ["not", "-",  ">", "<=>", ">=", "and", "is"]
```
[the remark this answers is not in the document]
macc125: Fixed syntax error with `<=` in c9a6928a949f9e78d048b1bdc65a1d7547e457ea

src/Adder/Lang/Lexer.hs
```diff
@@ -40,7 +40,7 @@ adderLexSpec =
       Tok.opStart = Tok.opLetter adderLexSpec,
       Tok.opLetter = oneOf "+-*/%@<>&|^~:=!.",
       Tok.reservedOpNames =
-        []
+        ["not", "-",  ">", "<=", ">=", "and", "is"]
```
Tientuine, line 43: The non-symbolic operators (`not`, `and`, `is`) should be in the reservedNames list instead. Only symbolic operator names should be here. However, in the interest of time, I'll make the change when resolving conflicts during merge.

src/Adder/Lang/Parser.hs
```diff
@@ -6,6 +6,7 @@
  -  for Monty Python.
  -
  -  This module provides the grammatical specification for Adder.
+ Adding comment to establish my own branch
```
Tientuine, line 9: No need for this comment. I will remove when doing the merge.

src/Adder/Lang/Parser.hs
```diff
@@ -65,7 +66,11 @@ compoundStmt =
 simpleStmt :: IParser Statement
 simpleStmt =
   (choice . map try)
-    []
+    [
+    BreakStmt <$ reserved "BreakStmt"
+
+
+    ]
```
Tientuine, line 73: Looks good, though there is no need for the two empty lines.

src/Adder/Lang/Parser.hs
```diff
@@ -74,22 +79,44 @@ stmtList = undefined
...
-    ],
-    [ Infix (reservedOp "<" >> return (BinaryExpr Less)) AssocLeft
-    ],
-    [ Infix (reservedOp "==" >> return (BinaryExpr Equal)) AssocLeft
-    ],
-    [ Infix (reserved "is" >> return (BinaryExpr Is)) AssocLeft
-    ],
-    [Prefix (reserved "not" >> return (UnaryExpr Not))],
-    [Infix (reserved "and" >> return (BinaryExpr And)) AssocLeft]
+  [
```
Tientuine, line 82: Good enough. There will be some editing needed during merge, so I'll fix the formatting at that time.

src/Adder/Lang/Parser.hs
```diff
@@ -74,22 +79,44 @@ stmtList = undefined
...
+	[Prefix (reserved "not" >> return (UnaryExpr Not))],
+
+	-- and
+	[Infix (reserved "and" >> return (BinaryExpr And)) AssocLeft],
+
+	 -- or
+	[Infix (reserved "or" >> return (BinaryExpr Or)) AssocLeft]
   ]
 
+
```
Tientuine, line 119: Try to avoid inserting empty lines in the future, as they show up as git changes even though no code has really changed.

src/Adder/Lang/Syntax.hs
```diff
@@ -25,9 +25,15 @@ data Expression
...
 
+
+
 data UnaryOp
   = Not
   | Negative
+
+
+
+
```
Tientuine, line 36: Again, lots of empty lines here. We'll remove these during merge.

Tientuine (approved):
Close enough! The code is okay, but the formatting is still off, including lots of extraneous blank lines added in several places. I'll take care of this during the merge this time, though in the future you can take care to maintain consistent code style.

PR #24, opened 2022-11-29
Continue Statement

jstengle15:


src/Adder/Lang/Parser.hs
```diff
@@ -65,7 +65,9 @@ compoundStmt =
 simpleStmt :: IParser Statement
 simpleStmt =
   (choice . map try)
-    []
+    [ ContinueStmt
+        <$> (reserved "continue")
+        ]
```
Tientuine, line 70: This will not work because a `ContinueStmt` does not contain anything. See the `pass` statement for an example of how this can be re-written.

src/Adder/Lang/Syntax.hs
```diff
@@ -16,7 +16,7 @@ newtype Program
 
 -- TODO Build out the abstract syntax for Adder statements
 data Statement
-  = NoStmt -- TODO Remove this after adding legit data constructors
+  = ContinueStmt Expression
```
Tientuine, line 19: A `ContinueStmt` is just the keyword `continue`, so there will not be any expression to store in this node. Instead, follow the same approach used for the `PassStmt`.

Tientuine (changes requested):
With a few fixes, the syntax part will be done, but you will still need to implement the interpeter piece later (once loops are implemented).

src/Adder/Lang/Lexer.hs
```diff
@@ -34,7 +34,7 @@ adderLexSpec =
       -- TODO Define the reserved names/keywords for the Adder language
       -- See https://docs.python.org/3/reference/lexical_analysis.html#keywords
       Tok.reservedNames =
-        [],
+        ["continue"],
```
Tientuine, line 37: Done.

src/Adder/Lang/Parser.hs
```diff
@@ -64,8 +64,9 @@ compoundStmt =
 -- See https://docs.python.org/3/reference/simple_stmts.html#grammar-token-python-grammar-simple_stmt
 simpleStmt :: IParser Statement
 simpleStmt =
-  (choice . map try)
-    []
+  (choice . map try) [ 
+      (reserved "continue" >> return ContinueStmt)  
+  ]
```
Tientuine, line 69: Okay.

src/Adder/Lang/Syntax.hs
```diff
@@ -16,7 +16,7 @@ newtype Program
 
 -- TODO Build out the abstract syntax for Adder statements
 data Statement
-  = NoStmt -- TODO Remove this after adding legit data constructors
+  = ContinueStmt
```
Tientuine, line 19: Good.

Tientuine (approved):
Looks ready for merge.

PR #25, opened 2022-11-29
Added support for addition and concatenation

RichieConsiglio:


Tientuine:
@RichieConsiglio It looks like all the changes that were going to be in this PR ultimately ended up in other branches, so we'll just close this one out without merging. I recommend that you delete your branch and then create a new one off the current `for-merging` branch before starting a new task.

PR #26, opened 2022-11-29
Smith work, Atom creation

Stubeans:
Atom syntax code


src/Adder/Lang/Lexer.hs
```diff
@@ -34,7 +34,7 @@ adderLexSpec =
       -- TODO Define the reserved names/keywords for the Adder language
       -- See https://docs.python.org/3/reference/lexical_analysis.html#keywords
       Tok.reservedNames =
-        [],
+        ["yield", "from"],
```
Tientuine, line 37: Fine to have these here, but we will probably not implement these kinds of statements in the interpreter.

Tientuine (changes requested):
We discussed in person in class, so I won't say much more here unless you have new questions that come up.

src/Adder/Lang/Lexer.hs
```diff
@@ -34,7 +34,7 @@ adderLexSpec =
       -- TODO Define the reserved names/keywords for the Adder language
       -- See https://docs.python.org/3/reference/lexical_analysis.html#keywords
       Tok.reservedNames =
-        ["or", "and", "is", "not", "in", "not in", "is not"],
+        ["yield", "from", "or", "and", "is", "not", "in", "not in", "is not"],
```
Tientuine, line 37: Let's not worry about `yield`, that's one of the language features that we said we would *not* try to implement.

app/Main.hs
```diff
@@ -7,6 +7,7 @@
  -  This module provides routines for executables based on HOPL languages.
  -
  -  Author: Matthew A Johnson
+ -  Test Line
```
Tientuine, line 10: Get rid of this test line, only useful changes should be included!

src/Adder/Defs.hs
```diff
@@ -19,4 +19,4 @@ type Source = String
 type Reference = Int
 
 -- Interpreter takes source code and produces a value (of some type)
-type Interpreter a = Source -> Either ParseError a
+type Interpreter a = Source -> Either ParseError a
```
Tientuine, line 22: Undo your change to this file - somehow you removed the trailing newline, and there's no need for that change.

src/Adder/Lang/Parser.hs
```diff
@@ -108,12 +108,56 @@ table =
     [Infix (reserved "or" >> return (BinaryExpr Or)) AssocLeft]
   ]
 
+atom :: IParser Expression
+atom = AtomExp <$> Id
```
Tientuine, line 112: This is incorrect, but I believe we already discussed this and you have an un-pushed correction for this on your computer.

Tientuine (changes requested):
Almost there... I believe you might already have a complete version that's waiting to be pushed, so I will keep an eye out for that.

Tientuine:
@Stubeans This piece - identifiers - is one of the crucial pieces remaining in order for us to test parsing a variety of simple programs. Let's get it done! Please let me know if you need help to wrap this one up.

Stubeans:
@Tientuine I believe I fixed the issues and fixed the Merge Conflicts correctly.

Tientuine (approved):

PR #27, opened 2022-11-29
Literals

cole-wahlers1:


src/Adder/Lang/Syntax.hs
```diff
@@ -23,6 +23,9 @@ data Statement
 data Expression
   = UnaryExpr UnaryOp Expression
   | BinaryExpr BinaryOp Expression Expression
+  | intLiteralExp Integer
+  | stringLiteralExp String
+  | floatLiteralExp Float
```
Tientuine, line 28: In Haskell, these must be capitalized since they are constructors.

src/Adder/Lang/Parser.hs
```diff
@@ -92,7 +92,13 @@ table =
...
 -- See https://docs.python.org/3/reference/expressions.html
 expression :: IParser Expression
-expression = buildExpressionParser table atom <?> "expression"
+expression = (choice . map try)
+    [ 
+      intLiteralExp <$> interger,
+      stringLiteralExp <$> string,
+      floatLiteralExp <$> float
+      
+    ]
```
Tientuine, line 101: Please restore this `expression` function back to its original form! We are relying on the `buildExpressionParser` function to do a lot of heavy lifting for us.

Take your code and make that the implementation of the `atom` function instead.

Tientuine (changes requested):
A few changes needed (see below), but it should not take much more effort to get this completed.

Tientuine (approved):

PR #30, opened 2022-11-29
Teixeira work

Tex1001:


src/Adder/Lang/Parser.hs
```diff
@@ -74,20 +74,29 @@ stmtList = undefined
 -- See https://docs.python.org/3/reference/expressions.html#operator-precedence
 table :: [[Operator String () (IndentT Identity) Expression]]
 table =
-  [ [Prefix (reservedOp "-" >> return (UnaryExpr Negative))],
-    [Infix (reservedOp "**" >> return (BinaryExpr Power)) AssocRight],
+  [ [ Prefix (reservedOp "-" >> return (UnaryExpr Negative))
+    ],
+    [ Prefix (reservedOp "+" >> return (UnaryExpr Positive))
+    ],
```
Tientuine, line 80: So close! Put these two operators in the same "row" of table (i.e., in the same sublist) to indicate that they have equal precedence.

src/Adder/Lang/Parser.hs
```diff
@@ -74,20 +74,29 @@ stmtList = undefined
...
+    [ Infix (reservedOp "**" >> return (BinaryExpr Power)) AssocRight
+    ],
     [ Infix (reservedOp "*" >> return (BinaryExpr Times)) AssocLeft
     ],
     [ Infix (reservedOp "+" >> return (BinaryExpr Plus)) AssocLeft
     ],
+    [ Infix (reservedOp "-" >> return (BinaryExpr Minus)) AssocLeft
+    ],
```
Tientuine, line 88: So close! As above with negative/positive, put this operator in the same row of the table with the addition operator `+`.

src/Adder/Lang/Parser.hs
```diff
@@ -74,20 +74,29 @@ stmtList = undefined
...
     [ Infix (reservedOp "<" >> return (BinaryExpr Less)) AssocLeft
     ],
+    [ Infix (reservedOp ">" >> return (BinaryExpr Great)) AssocLeft
+    ],
```
Tientuine, line 92: So close! As above with negative/positive, put this operator in the same row of the table with the less-than operator `<`.

src/Adder/Lang/Syntax.hs
```diff
@@ -26,17 +26,20 @@ data Expression
   deriving (Show)
 
 data UnaryOp
-  = Not
-  | Negative
+  = Negative
+  | Positive
+  | Not
```
Tientuine, line 31: Good, although I'm not sure why you changed the order here. The order of the constructors in this file does not have any impact on the operator precedence, that's handled entirely by the order of the rows in the `table`.

src/Adder/Lang/Syntax.hs
```diff
@@ -26,17 +26,20 @@ data Expression
...
 
 -- TODO Build out the abstract syntax for more binary operations
 data BinaryOp
   = Plus
+  | Minus
   | Times
   | Power
   | Equal
   | Less
+  | Great
```
Tientuine, line 42: Good.

Tientuine (changes requested):
Very nearly done, just a few fixes needed to ensure correct operator precedence.

src/Adder/Lang/Parser.hs
```diff
@@ -74,18 +74,21 @@ stmtList = undefined
...
     ],
-    [ Infix (reservedOp "+" >> return (BinaryExpr Plus)) AssocLeft
+    [ Infix (reservedOp "//" >> return (BinaryExpr Int_Divide)) AssocLeft
     ],
-    [ Infix (reservedOp "-" >> return (BinaryExpr Minus)) AssocLeft
+    ]
+    [[ Infix (reservedOp "+" >> return (BinaryExpr Plus)) AssocLeft
     ],
+    [ Infix (reservedOp "-" >> return (BinaryExpr Minus)) AssocLeft
+    ]],
```
Tientuine, line 91: Too many brackets now. The outermost set of bracket delimits the `table` as a whole. Each set of inner brackets denote a row, and all rules for parsing operators should be placed in a given row, comma-separated but not inside another set of brackets. In other words, the brackets should only go two-deep.

src/Adder/Lang/Syntax.hs
```diff
@@ -36,6 +36,7 @@ data BinaryOp
   = Plus
   | Minus
   | Times
+  | Int_Divide
```
Tientuine, line 39: Someone else already added this change, with a different name. When we merge, we'll only keep one of these names.

In the future, try to make sure that only one person commits a given change, to avoid more conflicts.

Tientuine (changes requested):
Many of your changes here are also included in some other branches. When they differ, we will only be able to keep one version. Please coordinate with others to help ensure that each change is only committed by one person, to avoid more conflicts.

Tientuine (approved):
A few conflicts/inconsistencies with changes others also pushed, but best for fix during merge and move forward.

PR #47, opened 2022-12-06
Added Suite

IanMarsh1:


src/Adder/Lang/Parser.hs
```diff
@@ -61,7 +61,7 @@ statement =
 
 -- See https://docs.python.org/3/reference/compound_stmts.html#grammar-token-python-grammar-suite
 suite :: IParser [Statement]
-suite = undefined
+suite = block statement
```
Tientuine, line 64: Okay, so actually this is not yet done. What you have so far is correct in part, but if you look at the reference link above, a `suite` can be either an indented block of statements _or_ it can be a statement list. So, here you need to account for both options, but using either the `(<|>)` operator that appears in the `atom` implementation at the bottom or the `(choice . map try) [...]` style that appears in some of the other functions.

Tientuine (changes requested):
Almost, one more change needed (see below).

Tientuine:
Unfortunately, I have to close this PR without merging because your branch now has unrelated changes (the Interp changes). I'll make a new branch called `Marsh-suite` from the previous commit of your branch, and then you can make a new PR for that branch and make the final change to `suite` on there.

PR #49, opened 2022-12-06
Interpreter support for binary operations

Stubeans:
Interp.hs implentation of Binary Operations

### DIFF
--- a/src/Adder/Interp.hs
+++ b/src/Adder/Interp.hs
@@ -76,10 +76,10 @@ type Answer = (ExpVal, Store)
 valueOf :: Expression -> Environment -> Store -> Answer
 valueOf _ env st0 = undefined
 -- Binary Operation
-valueOf (BinaryExp op exp1 exp2) env st0 = valueOfBop op val1 val2 st2
+valueOf (BinaryExp op exp1 exp2) env st0 = valueOfBop op val1 val2
   where
-    val1 st1 = valueOf(exp1) env st0
-    val2 st2 = valueOf(exp2) env st1
+    (val1, st1) = valueOf exp1  env st0
+    (val2, st2) = valueOf exp2 env st1
 
 -- Don't forget about free store
 
@@ -91,7 +91,7 @@ valueOf (BinaryExp op exp1 exp2) env st0 = valueOfBop op val1 val2 st2
 
 valueOfBop :: BinaryOp -> ExpVal -> ExpVal -> ExpVal
 valueOfBop op val1 val2 = case op of
-  _ -> undefined
+  _ -> error "unimplemented binary operation"
   --Code here
 
 --valueOf :: Return -> Environment -> Store -> Answer
--- a/src/Adder/Interp.hs
+++ b/src/Adder/Interp.hs
@@ -76,11 +76,10 @@ type Answer = (ExpVal, Store)
 valueOf :: Expression -> Environment -> Store -> Answer
 valueOf _ env st0 = undefined
 -- Binary Operation
-valueOf (BinaryExp op exp₁ exp₂) st0 = valueOfBop op1 val₁ val₂ st2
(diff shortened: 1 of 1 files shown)


src/Adder/Interp.hs
```diff
@@ -74,13 +75,24 @@ type Answer = (ExpVal, Store)
 -- TODO Implement the semantics for each kind of Adder expression
 valueOf :: Expression -> Environment -> Store -> Answer
 valueOf _ env st0 = undefined
+-- Binary Operation
+valueOf (BinaryExp op exp1 exp2) env st0 = valueOfBop op val1 val2 st2
+  where
+    val1 st1 = valueOf(exp1) env st0
+    val2 st2 = valueOf(exp2) env st1
```
Tientuine, line 82: You need parentheses and comma on the left-hand sides here because `valueOf` returns an `Answer`, which in this code is just a pair of `ExpVal` and `Store` (see Line 71).

src/Adder/Interp.hs
```diff
@@ -74,13 +75,24 @@ type Answer = (ExpVal, Store)
...
 --valueOF :: assignmentExpr ->  ??
 
 {- Auxiliary functions -}
 -- TODO Implement any helper functions needed to simplify the design of the
 -- interpreter (e.g., the applyProcedure helper function).
 
+valueOfBop :: BinaryOp -> ExpVal -> ExpVal -> ExpVal
+valueOfBop op val1 val2 = case op of
+  _ -> undefined
+  --Code here
```
Tientuine, line 95: Looks good. Maybe on Line 94 it would be helpful to others if you use something like `error "unimplemented binary operation"`, similar to what I mentioned in class earlier.

Tientuine (changes requested):
Needs a few very small adjustments, but then this will be ready for approval and merge.

src/Adder/Interp.hs
Tientuine: You also don't need the parens around `exp1` and `exp2`. They're not wrong, just unnecessary.

src/Adder/Interp.hs
```diff
@@ -24,6 +24,7 @@ import Adder.Lang.Syntax (Expression (..), Program (..), Statement (..))
 import Adder.Store (Store, deref, emptyStore, newref, setref)
 import Data.Either (fromRight)
 import Prelude hiding (exp)
+import GHC.Base (undefined)
```
Tientuine, line 27: Not be necessary, did VSCode suggest that you add this?

src/Adder/Interp.hs
```diff
@@ -74,13 +75,24 @@ type Answer = (ExpVal, Store)
 -- TODO Implement the semantics for each kind of Adder expression
 valueOf :: Expression -> Environment -> Store -> Answer
 valueOf _ env st0 = undefined
+-- Binary Operation
+valueOf (BinaryExp op exp1 exp2) env st0 = valueOfBop op val1 val2
+  where
+    (val1, st1) = valueOf exp1  env st0
+    (val2, st2) = valueOf exp2 env st1
+
+-- Don't forget about free store
```
Tientuine, line 84: Looks good. Needs to replace or be above the preceding line, but we can make that change later.

src/Adder/Interp.hs
```diff
@@ -74,13 +75,24 @@ type Answer = (ExpVal, Store)
...
 --valueOF :: assignmentExpr ->  ??
 
 {- Auxiliary functions -}
 -- TODO Implement any helper functions needed to simplify the design of the
 -- interpreter (e.g., the applyProcedure helper function).
 
+valueOfBop :: BinaryOp -> ExpVal -> ExpVal -> ExpVal
+valueOfBop op val1 val2 = case op of
+  _ -> error "unimplemented binary operation"
+  --Code here
```
Tientuine, line 95: Looks good.

Tientuine (approved):
Good to go.

PR #52, opened 2022-12-07
Marsh suite

IanMarsh1:
Sorry about that last push did it by accident. Also kind of shooting from the hip with this. I understand that I am trying to return [Statement], and stmtList is just a Statement. I think that adding block before it makes stmtList return a list, but I am still a little confused.

### DIFF
--- a/src/Adder/Lang/Parser.hs
+++ b/src/Adder/Lang/Parser.hs
@@ -61,7 +61,7 @@ statement =
 
 -- See https://docs.python.org/3/reference/compound_stmts.html#grammar-token-python-grammar-suite
 suite :: IParser [Statement]
-suite = undefined
+suite = block statement
 
 -- See https://docs.python.org/3/reference/compound_stmts.html#grammar-token-python-grammar-compound_stmt
 compoundStmt :: IParser Statement
--- a/src/Adder/Lang/Parser.hs
+++ b/src/Adder/Lang/Parser.hs
@@ -61,7 +61,7 @@ statement =
 
 -- See https://docs.python.org/3/reference/compound_stmts.html#grammar-token-python-grammar-suite
 suite :: IParser [Statement]
-suite = block statement
+suite = block statement <|> block stmtList
 
 -- See https://docs.python.org/3/reference/compound_stmts.html#grammar-token-python-grammar-compound_stmt
 compoundStmt :: IParser Statement


src/Adder/Lang/Parser.hs
```diff
@@ -61,7 +61,7 @@ statement =
 
 -- See https://docs.python.org/3/reference/compound_stmts.html#grammar-token-python-grammar-suite
 suite :: IParser [Statement]
-suite = undefined
+suite = block statement <|> block stmtList
```
Tientuine, line 64: Getting warmer... if you look carefully at the desired EBNF rule, you'll see that, unlike the block of `statement`s the `stmt_list` part is _not_ indented, it's just a stmt_list

> suite         ::=  [stmt_list](https://docs.python.org/3/reference/compound_stmts.html#grammar-token-python-grammar-stmt_list) NEWLINE | NEWLINE INDENT [statement](https://docs.python.org/3/reference/compound_stmts.html#grammar-token-python-grammar-statement)+ DEDENT

Tientuine (changes requested):
Teeny change left before approval.

Tientuine:
Good question, let's break it down a little at a time:

- Your `suite` function needs to be a parser for a list of `Statement`s (i.e., returns a `[Statement]` wrapped in an `IParser`).
- When you write the rule `block statement`, you're building a parser for a list of `Statement`s (because `statement` is a parser for a single `Statement` and `block` converts that into a parser for `[Statement]`).
- You also need a rule that effectively says that a `suite` can also be a `StmtList`, but the `stmtList` function gives you a parser for a `Statement` (i.e., returns a `Statement` wrapped in an `IParser`), which is the wrong type since you need `[Statement]`.
- However, the `StmtList` kind of statement does contain a `[Statement]`, so what you need to do is get the `[Statement]` from inside the `StmtList` and return it.

Here's a simple, but kind of analogous, Haskell example showing how to get the data encapsulated in one type so that we can pass it along. In this example, the `IO` type is taking the place of `IParser`.

```haskell
-- Imagine a data type that can encapsulate different kinds of values, includes itself
data Result = AResult Int | ManyResults [Result]

-- It can sometimes be helpful to have a function that can extract an encapsulated value
getResults :: Result -> [Result]; getResults r = case r of (ManyResults rs) -> rs; _ -> [r]

-- Now suppose we have a way of generating an instance of our data type
readResults :: IO Result
readResults = (ManyResults . map (AResult . read) . words) <$> getLine

-- But perhaps we also need a way to generate something like what's encapsulated in our type
-- One way we can do so is to extract the encapsulated data using our helper function
readResultList :: IO [Result]
readResultList = getResults <$> readResults
```

IanMarsh1:
A bit lost in the sauce. Can you tell me if I am on the right path or not?

src/Adder/Lang/Parser.hs
```diff
@@ -58,10 +58,20 @@ statement =
     [ compoundStmt,
       stmtList
     ]
+-- this gives the option to return one statement or a list of statements
+data Statement = AResult Statement | ManyResults [Statement]
+
```
Tientuine, line 63: My example was meant as a near analogy, not as code you needed to copy into here directly.

It turns out we already have a type named `Statement` defined in **Syntax.hs**. So you don't need this here at all.

src/Adder/Lang/Parser.hs
```diff
@@ -58,10 +58,20 @@ statement =
...
+-- take many results are return a list of statements if it is anything else return list
+getResults :: Statement -> [Statement] 
+getResults r = case r of (ManyResults rs) -> rs; _ -> [r]
+
```
Tientuine, line 67: This is the right idea, except that you should be matching on `(StmtList rs)`, which is the kind of `Statement` that contains the list-of-statements (rather than `ManyResults`, which is actually unnecessary here). Also, a better name might be `getStatements` in this case, given the context we're working in.

src/Adder/Lang/Parser.hs
```diff
@@ -58,10 +58,20 @@ statement =
...
+-- wrap the results in the parser
+readResultList :: IParser [Statement]
+readResultList = getResults <$> StmtList
```
Tientuine, line 70: Again, as with the data type you defined above, we don't actually need this new function, because `suite` is already the function that serves the same role as this one from my example. You could use it that way you are currently doing it, but it's just as easy to simply put the `getResults <$> stmtList` (or `getStatements <$> stmtList` if you rename it) right into the definition of `suite`.

src/Adder/Lang/Parser.hs
```diff
@@ -58,10 +58,20 @@ statement =
...
 
 -- See https://docs.python.org/3/reference/compound_stmts.html#grammar-token-python-grammar-suite
 suite :: IParser [Statement]
-suite = block statement <|> block stmtList
+suite = block statement <|> readResultList stmtList -- I have no idea what i am doing i am going to go to office hours tmmr
```
Tientuine, line 74: See my comment above regarding your `readResultList` function. It will work if you make a separate helper function, but in this case that function is so short it's probably just as easy (and arguably easier to read) if you just put the code right here.

Tientuine (changes requested):
A good attempt, but not quite there yet. See my per-line comments below for details.

PR #53, opened 2022-12-07
Boolean Literals

MaxEnglish:
I'm not sure if I implemented this correctly, please don't curb your criticism. Thanks

### DIFF
--- a/src/Adder/Lang/Syntax.hs
+++ b/src/Adder/Lang/Syntax.hs
@@ -37,6 +37,7 @@ data Expression
   | IntLiteralExp Integer
   | StringLiteralExp String
   | FloatLiteralExp Float
+  | BoolLiteralExp Boolean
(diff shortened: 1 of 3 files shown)


src/Adder/Lang/Syntax.hs
```diff
@@ -37,6 +37,7 @@ data Expression
   | IntLiteralExp Integer
   | StringLiteralExp String
   | FloatLiteralExp Float
+  | BoolLiteralExp Boolean
```
Tientuine, line 40: The name of the Boolean type in Haskell is simply `Bool`.

Tientuine (changes requested):
Just one small correction needed (see below for Syntax.hs) and then this PR can be approved and merged.

Tientuine (approved):
Looks good.

Issue #43, opened 2022-12-05
Interpret unary operations


Tientuine:
@Marist-CMPT331-TOPL/students Depends on #42 

Add cases to `valueOfUop` for each of the valid combinations of unary operators and operand. Note: Any combinations not covered by what you implement will simply be `undefined`.

- [ ] **Negative** - valid only for integer and float values, flip the sign
- [ ] **Positive** - valid only for integer and float values, do nothing but pass the value through unchanged
- [ ] **Logical Negation** - initially only for Boolean values, flip the truth value as usual

Note: The implementation of `not` here will serve as an initial implementation. Time permitting, we may tweak this later to make it more faithful to how Python does things, where we determine the "truthiness" of the operand and then return an appropriate Boolean.

PR #54, opened 2022-12-08
interp support for unaryop

JGargana1:


src/Adder/Interp.hs
```diff
@@ -59,7 +59,10 @@ type Answer = (ExpVal, Store)
 -- TODO Implement the semantics for each kind of Adder expression
 valueOf :: Expression -> Environment -> Store -> Answer
 valueOf _ env st0 = undefined
+valueOf UnaryExpr = valueOfUop ()
 
 {- Auxiliary functions -}
 -- TODO Implement any helper functions needed to simplify the design of the
 -- interpreter (e.g., the applyProcedure helper function).
+valueOfUop :: UnaryOp -> Expression -> Expression -> return (UnaryExpr)
```
Tientuine, line 67: I recommend replacing the `Expression`s in the type of this function with `ExpVal`s. This will make it easier for those students working on the unary operations to avoid writing conflicting code. What this means for you is that instead of passing in two expressions and having to evaluate them here to produce the values, you'll evaluate the two expressions in `valueOf` above, and then just feed the resulting values into this function.

Also, the return type here should be `ExpVal`. This is part of the interpreter, and we don't want to build another syntax node (which is what `UnaryExp` is). Instead we want to compute and return an actual value. (Furthermore, the `return` function in Haskell cannot be used as part of the function's _type_.)

So, in summary, the type for this function should be `valueOfUop :: UnaryOp -> ExpVal -> ExpVal -> ExpVal`.

P.S. I know that the version seen in the LETREC language implementation does it differently, but (a) that was written by one person not several collaborating and (b) if I were to do it over again I would implement it the way I'm suggesting to you here.

src/Adder/Interp.hs
```diff
@@ -59,7 +59,10 @@ type Answer = (ExpVal, Store)
 -- TODO Implement the semantics for each kind of Adder expression
 valueOf :: Expression -> Environment -> Store -> Answer
 valueOf _ env st0 = undefined
+valueOf UnaryExpr = valueOfUop ()
```
Tientuine, line 62: Note the type of `valueOf` shown on line 60. This function takes three parameters - an Expression, an Environment, and a Store - and then returns an Answer (which is just a pair containing ExpVal and Store). You need to reflect all three parameters in your case.

Also, the `UnaryExp` constructor has two parameters itself, so you need to account for those when you write the pattern here on the left-hand side.

As a similar example, consider the rule for `AssignExp` from the IMPLICIT_REFS language (not the same kind of thing behaviorally, but very similar structurally).

```haskell
valueOf (AssignExp var exp) env st0 = ..`
```

Here, `AssignExp` is a constructor with two parameters, like your `UnaryOp`. And we can also see parameter names `env` and `st0` for the Environment and Store, respectively.

src/Adder/Interp.hs
```diff
@@ -59,7 +59,10 @@ type Answer = (ExpVal, Store)
...
 
 {- Auxiliary functions -}
 -- TODO Implement any helper functions needed to simplify the design of the
 -- interpreter (e.g., the applyProcedure helper function).
+valueOfUop :: UnaryOp -> Expression -> Expression -> return (UnaryExpr)
+valueOfUop _env st0 = undefined
```
Tientuine, line 68: Your `valueOfUop` function does not take in an Environment nor a Store, so the parameter names `env` and `st0` don't really make sense here. Instead you should provides names the `UnaryOp` and the two `ExpVal`s that will be passed into this function.

Also, on the right-hand side, there is another small change that would make it more straightforward for others to add the various operations themselves. You can add a `case`-`of` expression with a default case that is undefined. Then other students can just insert new lines rather than one of them having to modify your existing line of code to replace the `undefined`.

Here's an example of this kind of thing. Suppose we have a data type that represents the four traditional suits used in card games: `Hearts`, `Diamonds`, `Clubs`, `Spades`. Then imagine that we are implementing a game in which something different happens depending on which suit the card is, but different programmers will be implementing the logic for different suits. I might set up the function like this:

```haskell
playCard :: Rank -> Suit -> Outcome
playCard r s = case s of
    -- other programmers can now just insert their new lines here
    _ -> undefined  -- default case catches all suits that don't have a case preceding this
```

src/Adder/Interp.hs
Tientuine, line 62: You will also need to change the right-hand side to implement this correctly. The steps are basically this:

1. Evaluate the operand expression that's contained in the `UnaryExpr`. This will produce an Answer; i.e., a value and a new state of the free store.
2. Call `valueOfUop` and pass in the operator and the value you just got from evaluating the inner expression.
3. Make sure to return an Answer (i.e., pair) containing both the value you get from `valueOfUop` and the new free store (the same one that resulted from evaluating the inner expression, since no other change in state can have happened after that).

Tientuine (changes requested):
A reasonable start, but some changes are needed as described for each change below.

src/Adder/Interp.hs
```diff
@@ -59,10 +59,14 @@ type Answer = (ExpVal, Store)
...
+  diff -> NumVal (n1 - n2)
+  plus -> NumVal (n1 + n2)
+
+
 
 {- Auxiliary functions -}
 -- TODO Implement any helper functions needed to simplify the design of the
 -- interpreter (e.g., the applyProcedure helper function).
-valueOfUop :: UnaryOp -> Expression -> Expression -> return (UnaryExpr)
+valueOfUop :: UnaryOp -> ExpVal ->   ExpVal -> ExpVal
```
Tientuine, line 71: Good, though you have an extra space before the second `ExpVal` in the function type.

Also, see the `valueOfBop` function in Stuart's branch for an alternative way to write the undefined case here. The approach used there will make things go more smoothly (less chance of merge conflicts) when others add the actual unary operations.

src/Adder/Interp.hs
```diff
@@ -59,10 +59,14 @@ type Answer = (ExpVal, Store)
 -- TODO Implement the semantics for each kind of Adder expression
 valueOf :: Expression -> Environment -> Store -> Answer
 valueOf _ env st0 = undefined
-valueOf UnaryExpr = valueOfUop ()
+valueOf (UnaryExpr UnaryOp expression) env st0 = case op of
```
Tientuine, line 62: The left-hand side of this definition is close, but you can't say `UnaryOp` here because that's a type name. Instead, just pick a variable name to represent the `UnaryOp` that comes into this function... you can choose any name, but something short and simple like `op` is probably a good idea.

src/Adder/Interp.hs
```diff
@@ -59,10 +59,14 @@ type Answer = (ExpVal, Store)
...
+  diff -> NumVal (n1 - n2)
+  plus -> NumVal (n1 + n2)
+
+
```
Tientuine, line 66: Here, you don't need to actually implement any unary operations - those will be implemented below in the `valueOfUop` function you've started and are covered by issue #43 which Mac assigned himself to.

Instead, what you need to do in your implementation here is...

1. Evaluate the `expression`, which gives you a value (let's call it `val1`) and a new state (let's call it `st1).
2. Call your `valueOfUop` helper function, passing in the operator and `val1`, and this will give you back a new value let's call it `val2`) that will be the result of computing the unary operation.
3. Finally, this function needs to return the final value `val2` along with the changed state `st1` from when you evaluated the expression.

Tientuine (changes requested):
Good progress, but there are still some adjustments needed before it's ready.

PR #55, opened 2022-12-08
Interpret the value of a variable

Tones55:
not sure why in interp i have an error, Parse error in pattern: varExp

### DIFF
--- a/src/Adder/Interp.hs
+++ b/src/Adder/Interp.hs
@@ -20,7 +20,7 @@ import Adder.DataStructures (DenVal, Environment, ExpVal (..), Function (..))
 import Adder.Defs (Source)
 import Adder.Environment (Env (..))
 import Adder.Lang.Parser (ParseError, parseFile, parseInteractive)
-import Adder.Lang.Syntax (Expression (..), Program (..), Statement (..))
+import Adder.Lang.Syntax (Expression (..), Program (..), Statement (..), Atom (..))
 import Adder.Store (Store, deref, emptyStore, newref, setref)
 import Data.Either (fromRight)
 import Prelude hiding (exp)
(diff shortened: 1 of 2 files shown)


src/Adder/Interp.hs
```diff
@@ -73,7 +73,11 @@ type Answer = (ExpVal, Store)
 
 -- TODO Implement the semantics for each kind of Adder expression
 valueOf :: Expression -> Environment -> Store -> Answer
-valueOf _ env st0 = undefined
+valueOf (varExp x) env st0 = Answer (deref addr st0) st0
```
Tientuine, line 76: More more fix needed... To simplify the syntax a bit, in this interpreter implementation I defined an `Answer` to simply be another name for a pair of value and store, rather than its own new data type (see line 70). For this reason, there is no `Answer` constructor, and instead you can create the pair you need just by writing `(deref addr st0), st0)`.

Tientuine (changes requested):
A few minor fixes needed, but you're on the right track.

Tones55:
I made the changes you suggested

src/Adder/Lang/Parser.hs
```diff
@@ -148,6 +148,9 @@ assignmentExpr = expression -- For now, assignment expression only needs to be a
 --  [<$> identifier
 --    <*> (reservedOp "=" >> Expression)]
 
+varExpr :: IParser Expression
+varExpr = expression
+
```
Tientuine, line 153: It doesn't look like this gets used anywhere, so it's unnecessary, but also doesn't cause any problem, so we'll leave it for now.

Tientuine (approved):
The interpreter logic looks correct now. The new function you added to the Parser isn't really needed, but that's fine.

PR #56, opened 2022-12-08
Varca work

rvarca22:
Attempted resultOf IfStmt and IfExp in interpreter


rvarca22:
Issues at the moment:
- interpreter doesn't accept my 'where' clauses 
- if I delete the where,  interpreter doesn't accept my '=' sign...

Tientuine:
> Issues at the moment:
> 
>     * interpreter doesn't accept my 'where' clauses
> 
>     * if I delete the where,  interpreter doesn't accept my '=' sign...

See my review comments, they explain the issue (missing else). This does _not_ mean that you have to add support for else in the Adder language yet, it just means that in your Haskell code you can't leave out the else.

Tientuine (changes requested):

src/Adder/Interp.hs
```diff
@@ -51,20 +51,20 @@ resultOfProgram _ env st0 = undefined
 resultOf :: Statement -> Environment -> Store -> IO Store
 resultOf _ env st0 = undefined
 
+resultOf (IfStmt test conseq) p st = if q then st2 
+  where
+    Answer (BoolVal q) st1 = valueOf test p st
+    st2 = resultOf conseq p st1
+
+
```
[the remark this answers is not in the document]
rvarca22: Resolved

src/Adder/Interp.hs
```diff
@@ -75,6 +75,13 @@ type Answer = (ExpVal, Store)
 valueOf :: Expression -> Environment -> Store -> Answer
 valueOf _ env st0 = undefined
 
+valueOf (IfExp exp1 exp2 exp3) p st = valueOf exp' p st1 
+  where 
+    Answer q st1  = valueOf exp1 p st 
+    exp' = case q of 
+      BoolVal True -> exp2
+      BoolVal False -> exp3
+
```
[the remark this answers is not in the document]
rvarca22: Resolved (commented out). I will try to implement IfExp in syntax and parser to match my interpreter

src/Adder/Interp.hs
```diff
@@ -51,20 +51,20 @@ resultOfProgram _ env st0 = undefined
 resultOf :: Statement -> Environment -> Store -> IO Store
 resultOf _ env st0 = undefined
 
+resultOf (IfStmt test conseq) p st = if q then st2 
+  where
+    Answer (BoolVal q) st1 = valueOf test p st
```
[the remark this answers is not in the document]
rvarca22: Resolved

src/Adder/Interp.hs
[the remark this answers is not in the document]
Tientuine: In the parser, be sure to include the if-expression in the `table` list. In the link to the Python docs, you can see where it appears in the operator precedence table.

src/Adder/Interp.hs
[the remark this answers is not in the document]
Tientuine: Did you push your branch? Because no new changes are showing up here on GitHub.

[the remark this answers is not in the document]
rvarca22: Forgot to make the commit, should be up now

Trent325 (commented):
Looks pretty good Ryan I think this is right

src/Adder/Lang/Syntax.hs
```diff
@@ -35,6 +35,7 @@ data Expression
   | IntLiteralExp Integer
   | StringLiteralExp String
   | FloatLiteralExp Float
+  | ConditionalExpr [Expression]
```
Tientuine, line 38: A conditional expression cannot have any number of parts, which is what a list-of-expressions allows for. It must have exactly three sub-expressions, so in this case it would be best to simply have each of those be separately declared here. (As suggested above, see the `IfExp` from the HOPL.LET language.)

src/Adder/Interp.hs
```diff
@@ -75,16 +75,25 @@ type Answer = (ExpVal, Store)
 valueOf :: Expression -> Environment -> Store -> Answer
 valueOf _ env st0 = undefined
 
+If Expression not yet implemented in Syntax and Parser 
+valueOf (IfExp exp1 exp2 exp3) env st0 = valueOf exp' env st1
```
Tientuine, line 79: This is almost correct, except for two details.

1. First, `IfExp` doesn't match anything in the abstract syntax. In Syntax.hs, you called this `ConditionalExpr`, which it think is a fine name, but you must use that same name here.
2. Then, on the right-hand side, you must return the `ExpVal` and `Store` as a pair, which means they should be comma-separated and wrapped in parentheses. For example, if I needed to take the return value of `f x` and pair that with some other value `y`, then I would write `(f x, y)`.

src/Adder/Interp.hs
```diff
@@ -75,16 +75,25 @@ type Answer = (ExpVal, Store)
 valueOf :: Expression -> Environment -> Store -> Answer
 valueOf _ env st0 = undefined
 
+If Expression not yet implemented in Syntax and Parser 
+valueOf (IfExp exp1 exp2 exp3) env st0 = valueOf exp' env st1 
+  where 
+    Answer q st1  = valueOf exp1 env st0
```
Tientuine, line 81: A related problem to the issue on the right-hand side of line 79, `valueOf` returns a pair of `ExpVal` and `Store`, so on the left-hand side here you must represent it as a pair, like `(q, st1)`.

src/Adder/Interp.hs
```diff
@@ -75,16 +75,25 @@ type Answer = (ExpVal, Store)
 valueOf :: Expression -> Environment -> Store -> Answer
 valueOf _ env st0 = undefined
 
+If Expression not yet implemented in Syntax and Parser
```
Tientuine, line 78: This looks like it's supposed to be a comment, in which case the line should start with `--`. Otherwise, remove this line because it is not valid Haskell code.

src/Adder/Interp.hs
```diff
@@ -75,16 +75,25 @@ type Answer = (ExpVal, Store)
...
 -- TODO Implement any helper functions needed to simplify the design of the
 -- interpreter (e.g., the applyProcedure helper function).
 
 
---valueOf :: Return -> Environment -> Store -> Answer
---valueOf (Return exp1) env store = env2 --Added Exp 1 into the parathenses
---Answer Return exp1 env = exp1         --Attempted to add the return statement 
+valueOf :: Return -> Environment -> Store -> Answer
+valueOf (Return exp1) env store = env2 --Added Exp 1 into the parathenses
+Answer Return exp1 env = exp1         --Attempted to add the return statement
```
Tientuine, line 96: I recommend keeping this commented out, because we can't really implement `ReturnExpr` until we have functions.

src/Adder/Lang/Parser.hs
```diff
@@ -132,14 +132,21 @@ table =
     ],
     [Prefix (reserved "not" >> return (UnaryExpr Not))],
     [Infix (reserved "and" >> return (BinaryExpr And)) AssocLeft],
-    [Infix (reserved "or" >> return (BinaryExpr Or)) AssocLeft]
+    [Infix (reserved "or" >> return (BinaryExpr Or)) AssocLeft],
+    [infix (reserved "if" >> return (conditionalExpr if)), AssocLeft]
```
Tientuine, line 136: `Infix` should be capitalized here, and you should remove the `if` after `conditionalExpr`.

src/Adder/Lang/Parser.hs
```diff
@@ -132,14 +132,21 @@ table =
...
 -- See https://docs.python.org/3/reference/expressions.html
 expression :: IParser Expression
 expression =
   buildExpressionParser table atom
     <?> "expression"
+    -- add conditionalExpr as a type of Expression - map data?
+
+conditionalExpr :: Iparser Expression -- RV: not sure how to implement a 'or_test'
+conditionalExpr = orTest [ reserved "if" or_test "else" expression]
+  --or_test  ::=  and_test | or_test "or" and_test
```
Tientuine, line 149: This function is a good idea, but I think that trying to match what's in the Python docs exactly is going to be more effort than it's worth. Instead, just simplify things by assuming that each of the three parts is just an `Expression`. So, if you do that, it will look very much like the `IfExp` from the HOPL.LET language.

In that case, the rule was `IfExp <$> (reserved "if" >> expression) <*> (reserved "then" >> expression) <*> (reserved"else" >> expression)`. In your case, it will be very similar, except the keywords and the order of the parts.

Tientuine (changes requested):
Incidentally, the `IfExp` from the HOPL languages is exactly a conditional expression, just that its syntax is a bit different.

For example, in LET we might have `if zero?(x) then y else z`. However, in Adder it would look like `y if x == 0 else z`. Also, in the abstract syntax of LET, the first sub-expression is the test, the second is the true-branch, and the third is the false-branch. However, in Adder the first sub-expression is the true-branch, the second is the test, and the third is still the false-branch. Despite these syntactic differences, in the interpreter they are evaluated in exactly the same way.

PR #58, opened 2022-12-08
dahir-work

bashirad:
Interp.hs => augmented assignment statement implementation


src/Adder/Interp.hs
```diff
@@ -49,7 +49,47 @@ resultOfProgram _ env st0 = undefined
...
+    rval3 = IntVal ( rval1 *  rval2)
+    st2 = setref address1 rval3 st1
+
+resultOf (AugmentedAssignmentStmt var AugDiv rhs) env st0 = return st2
+  where
+    address1 = applyEnv env var
+    IntVal (rval1) = deref address1 st0
+    (IntVal (rval2), st1) = valueOf rhs env st0
+    rval3 = IntVal (div rval1 rval2) -- floor and frominteger
+    st2 = setref address1 rval3 st1
```
Tientuine, line 92: This all looks fine for now, except that some of the formatting is off. Is your Haskell code formatting not working? If I recall correctly, you're using IntelliJ, right? See https://plugins.jetbrains.com/plugin/8258-intellij-haskell and https://tek.brick.do/setting-up-ormolu-for-intelli-j-idea-erqLxeMJXPxk for suggestions on setting up better support for Haskell, including the latest ormolu formatter.

src/Adder/Interp.hs
```diff
@@ -71,6 +111,7 @@ type Answer = (ExpVal, Store)
 
 {- semantic reductions for expressions -}
 
+
```
Tientuine, line 114: Try to avoid inserting non-code like this stray empty line. It just adds noise to the git change history.

Tientuine (changes requested):
The functionality is good (if not yet complete for all combinations of `ExpVal`s), but I would like you to make a few cosmetic changes before approving.

Tientuine (approved):
I think this looks correct, though we need to merge to properly test it. Stay tuned in case any fixes are needed.

PR #59, opened 2022-12-09
StmtList Interp

IanMarsh1:


Tientuine (approved):

Tientuine (changes requested):
The `_` pattern will match any `Statement`, so when we try to interpret a `StmtList` or `PassStmt`, we'll just get `undefined` instead of running the implementations you wrote. You'll need to move the default case down to below your new cases, so it only handles those statements that don't have a specific implementation.

IanMarsh1:
Done. I still want to try and add the helper function; I just won't be able to do it till after the final on Tuesday.

Tientuine (approved):

PR #60, opened 2022-12-09
if elif and else statements

MaxEnglish:
I tried the first approach you sent me in your email. I'm certain I didn't do it correctly though. Where online or in the textbook can I learn how to implement expressions in the interpreter? Let me know what I can change

### DIFF
--- a/src/Adder/Interp.hs
+++ b/src/Adder/Interp.hs
@@ -59,11 +59,35 @@ resultOf _ env st0 = undefined
 ---------------------------------------------
 -- env1 = env0
 {-
-resultOf (IfStmt test conseq) p st = if q then st2 else st3
+resultOf (IfStmt test conseq) p st = if q then st2
+  where
+    Answer (BoolVal q) st1 = valueOf test p st
+    st2 = resultOf conseq p st1
+-}
+
+{-
+resultOf (IfElseStmt test conseq) p st = if q then st2 else st3
   where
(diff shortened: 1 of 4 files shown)


Tientuine (commented):
A quick glance suggests that this is close to working, but not quite. Rather than trying to do all of these at once, start with just one (e.g., `IfElseStmt` and just compute the `resultOf` that. Once you get one of these correct, such as if-else, then the others should become a little easier.

I'll take a deeper look early tomorrow and post some more specific suggestions for how to improve it.

MaxEnglish:
Thanks

Tientuine (changes requested):
I've described the necessary changes for `IfElseStmt` below, so let's start with that.

Please make sure the code compiles before you continue on. Although you won't be able to test it interactively yet (because it may depend on a couple of other pieces that classmates are working on), it will most likely be correct if it compiles.

MaxEnglish:
[image: image.png]
This is how I've changed IfElseStmt based on your criticism. However, I'm
not sure how to "take the list-of-statements and wrap it in StmtList before
passing it to resultOf." I imagine it has to due with brackets which is why
I wrapped the exps in them but again I don't know if that's correct.

When I try to run the program, it fails in the parser. Below is a
screenshot of a portion of the error message. This somehow seems related to
the issue you outlined with the param needing to be a list of stmts.
[image: image.png]

PR #61, opened 2022-12-11
Green work

Trent325:
Attempted to add interrupter support for assignmentExp and assignmentStmt


src/Adder/Interp.hs
```diff
@@ -51,6 +51,19 @@ resultOfProgram _ env st0 = undefined
 resultOf :: Statement -> Environment -> Store -> IO Store
 resultOf _ env st0 = undefined
 
+resultOf (AssignmentStmt x body) env st0 = return rhs enva -- need to have a parameter name for the Store on the left side
+  where
+    (addr, st1) = newref body st0
+    env1 = extendEnv x env addr-- need to provide an address that the variable is mapped to
```
Tientuine, line 57: You have the arguments to `extendEnv` in the wrong order. The first two arguments should be the variable name and its value (in this case, the address), and the last argument should be the current environment.

src/Adder/Interp.hs
```diff
@@ -75,17 +88,20 @@ type Answer = (ExpVal, Store)
...
 valueOf _ env st0 = undefined
 
+ -- need to have a parameter name for the Store on the left side
+ -- on the right side, you need to just have a pair of value and store
+valueOf (AssignmentExpr x body) env st0 = rhs st2
+  where
+    (addr, st1) = newref v st0
+    env1 = extendEnv x env addr -- need to provide an address that the variable is mapped to
+    x = valueOf rhs env st0 -- need to pass in a Store parameter here also
+
```
Tientuine, line 98: See my comment(s) above for `resultOf` - the same suggestions apply here.

test/trenttest.add
```diff
@@ -0,0 +1,2 @@
+x = 5
+y = 'the'
```
Tientuine, line 2: Does our Adder implementation support single-quoted strings? You should check the Lexer to see.

Tientuine (changes requested):
Still some work to be done here, refer to the `AssignExp` from IMPLICIT_REFS and the `AssignStmt` from SIMPLE_STATEMENT to see other working implementations. It's important to understand the different parts of those examples before you can fix your code here. Then be sure to note any differences (e.g., how the `Answer` type is defined differently here), and implement accordingly.

src/Adder/Interp.hs
```diff
@@ -51,11 +51,11 @@ resultOfProgram _ env st0 = undefined
...
 resultOf _ env st0 = undefined
 
-resultOf (AssignmentStmt x body) env st0 = return rhs enva -- need to have a parameter name for the Store on the left side
+resultOf (AssignmentStmt var rhs) env st0 = return st2 -- need to have a parameter name for the Store on the left side
   where
-    (addr, st1) = newref body st0
-    env1 = extendEnv x env addr-- need to provide an address that the variable is mapped to
-    x = valueOf rhs env st 0  -- valueOf takes a Store parameter also
+    (addr, st1) = valueOf rhs env st0
+    env1 = extendEnv addr var env-- need to provide an address that the variable is mapped to
```
Tientuine, line 57: I think you skipped a step here. You evaluated the right-hand side, but this produces an `ExpVal` and a new store. To get the address, you need to allocate a new reference in the store, as you were doing originally.

src/Adder/Interp.hs
```diff
@@ -51,11 +51,11 @@ resultOfProgram _ env st0 = undefined
...
+    st2 = setref rhs addr st1  -- valueOf takes a Store parameter also
```
Tientuine, line 58: You don't really need this if you add the `newref` back in, since that will set the initial value in the free store. If you keep this, it won't cause an issue as long as you fix the type error - the `setref` function has type `Reference -> ExpVal -> Store -> Store`, meaning that the first argument should be the address and the second should be the value. (Note: `rhs is _not_ the value, it's the expression; you get the value by evaluating `rhs`).

Tientuine (changes requested):
Nearly there.

src/Adder/Interp.hs
```diff
@@ -53,7 +53,7 @@ resultOf _ env st0 = undefined
 
 resultOf (AssignmentStmt var rhs) env st0 = return st2 -- need to have a parameter name for the Store on the left side
   where
-    (addr, st1) = valueOf rhs env st0
+    (addr, st1) = newref rhs env st0
```
Tientuine, line 56: I think maybe you misunderstood me. You need to restore `newref` but not remove the `valueOf`. They serve two different purposes.

1. First you need to compute the value of the right-hand side expression, which you can do using `valueOf rhs env st0` exactly as you had. Except that `valueOf` returns a pair of expressed value and store, not reference and store. So it should look more like `(val1, st1)` on the left-hand side.
2. Putting `newref` back in is necessary because once you've computed the value of `rhs`, then you need to store that value in the free store, which is what `newref` does. Only, you need to pass in the expressed value and store that you got from `valueOf`, as in `newref val1 st1`. This will return a pair of reference and free store, but this will be the next free store after `st1`, so you'll have something like `(addr, st2)` here.

src/Adder/Interp.hs
```diff
@@ -92,7 +92,7 @@ valueOf _ env st0 = undefined
  -- on the right side, you need to just have a pair of value and store
 valueOf (AssignmentExpr var rhs) env st0 = rhs st1
   where
-    (addr, st1) = valueOf exp1 env st0
+    (addr, st1) = newref exp1 env st0
```
Tientuine, line 95: Same critique here as above. Once you get one working, the other should fall into place. But also you'll need to change the right-hand side on line 93, because you need to return the two values, ExpVal and Store, in a pairlike `(x, y)`. (Or course, you're not returning x and y, those are just placeholders in my example for the first and second values in the pair.)

Tientuine (changes requested):
A little closer each time, hopefully one more revision will do the trick!

PR #62, opened 2022-12-11
Interp literals

cole-wahlers1:


src/Adder/Interp.hs
```diff
@@ -59,6 +59,12 @@ type Answer = (ExpVal, Store)
 -- TODO Implement the semantics for each kind of Adder expression
 valueOf :: Expression -> Environment -> Store -> Answer
 valueOf _ env st0 = undefined
+-- String literal
+valueOf (StringLiteralExp var) ρ = applyEnv ρ var
+-- Integer literal
+valueOf (IntLiteralExp n) _ = NumVal n
+-- Float literal
+valueOf (FloatLiteralExp n) _ = Numval n
```
Tientuine, line 67: We have no `NumVal` constructor in the Adder implementation. Take a look at the definition of `ExpVal` in the Syntax.hs file - you can see that the constructor for integer values is called `IntVal` and the constructor for floating-point values is called `FloatVal`.

src/Adder/Interp.hs
```diff
@@ -59,6 +59,12 @@ type Answer = (ExpVal, Store)
 -- TODO Implement the semantics for each kind of Adder expression
 valueOf :: Expression -> Environment -> Store -> Answer
 valueOf _ env st0 = undefined
+-- String literal
+valueOf (StringLiteralExp var) ρ = applyEnv ρ var
```
Tientuine, line 63: A `StringLiteral` contains a String value, but it does not represent a variable name, so you should not expect to find it by looking it up in the environment. Instead you should construct and return an `StrVal`, similar to what you do for integers and floating-point numbers.

Tientuine (changes requested):
I don't understand why so many changes are showing up here. Did you copy and paste the latest code into your branch by hand, instead of merging the changes in using git?

Also, there are some errors in this PR that need to be fixed before approval.

src/Adder/Interp.hs
```diff
@@ -59,6 +59,13 @@ type Answer = (ExpVal, Store)
 -- TODO Implement the semantics for each kind of Adder expression
 valueOf :: Expression -> Environment -> Store -> Answer
 valueOf _ env st0 = undefined
+-- String literal
+valueOf (StringLiteralExp n) env = StrVal n
+-- Float literal
+valueOf (FloatLiteralExp n) _ = FloatVal n
+-- Integer literal
+valueOf (IntLiteralExp n) _ = IntVal n
+
```
Tientuine, line 68: Very close! However, the `valueOf` function returns an `Answer`, which here is just a pair of ExpVal and Store. You've got the ExpVal part, for example `(StrVal n)`, but you don't have the store part, which will just be the same free store that you had to start with because a literal value doesn't change anything in memory. In fact, the left-hand side of your definitions here are also missing the free store, which should appear as the third argument to the function, `st0`, as on line 61.

Tientuine (changes requested):

src/Adder/Interp.hs
```diff
@@ -59,6 +59,13 @@ type Answer = (ExpVal, Store)
 -- TODO Implement the semantics for each kind of Adder expression
 valueOf :: Expression -> Environment -> Store -> Answer
 valueOf _ env st0 = undefined
+-- String literal
+valueOf (StringLiteralExp n) st0 = StrVal n
+-- Float literal
+valueOf (FloatLiteralExp n) st0 = FloatVal n
+-- Integer literal
+valueOf (IntLiteralExp n) st0 = IntVal n
+
```
Tientuine, line 68: The type of `valueOf` is `Expression -> Environment -> Store -> Answer`. Unpacking that, the first argument to the function should be an `Expression`, which is covered by `(StringLiteralExp n)`. The second argument should be an `Environment`, so what you've called `st0` here will actually be an `Environment` even though you probably intended `st0` to be the `Store`. The problem is that even while you added `st0`, you also removed `env`.

The last part of the function type is the return type, `Answer`. On line 55, we can see that `Answer` is really just another name for a pair of `(ExpVal, Store)`. But on the right-hand side of your definition you only have `(StrVal n)`, which is the `ExpVal` part, but you're missing the `Store`. What you need is a pair like `(StrVal n, st0)`.

Although you've added the store `st0` to the left-hand side, you removed the environment, which is also needed. Again, the right-hand side is missing the store, `st0`.

Tientuine (changes requested):

PR #63, opened 2022-12-12
WhileStmt Work

DuncanMcLean:
I'm pretty confident this isn't 100% correct, please let me know what needs to be fixed or where I can find more information on it


src/Adder/Interp.hs
```diff
@@ -75,6 +75,10 @@ type Answer = (ExpVal, Store)
 valueOf :: Expression -> Environment -> Store -> Answer
 valueOf _ env st0 = undefined
 
+resultOf (WhileStmt test conseq) p st = while q st2
+  where
+    Answer (BoolVal q) st1 = valueOf test p st2
+
```
Tientuine, line 81: Since you are implementing a case of the `resultOf` function, this whole block should be moved up above to where the definition of `resultOf` currently is on line 52.

Tientuine (changes requested):
To See my more specific comments below for a description of what needs fixing and where to look for a similar example.

DuncanMcLean:
I made the updates you suggested, I looked at the CHECKED_STATEMENT language and tried to incorporate their implementation as you suggested.

src/Adder/Interp.hs
```diff
@@ -51,6 +51,15 @@ resultOfProgram _ env st0 = undefined
...
 resultOf _ env st0 = undefined
 
+resultOf (WhileStmt test conseq) p st0 = 
+  if q
+    then do 
+      st2 <- resultOf body p st1
+      resultOf stmt p st2
+    else return st1
+  where
+    (BoolVal q) st1 = valueOf test p st0
```
Tientuine, line 61: To make a pair with the `BoolVal q` and `st1`, you need to wrap in parentheses, comma-separated. For example, if I have an integer 5 and a string "Hello", then I could make a pair as `(5, "Hello")`.

src/Adder/Interp.hs
```diff
@@ -51,6 +51,15 @@ resultOfProgram _ env st0 = undefined
 resultOf :: Statement -> Environment -> Store -> IO Store
 resultOf _ env st0 = undefined
 
+resultOf (WhileStmt test conseq) p st0 = 
+  if q
+    then do 
+      st2 <- resultOf body p st1
+      resultOf stmt p st2
+    else return st1
```
Tientuine, line 59: The body of the loop will not be a single `Statement`, but rather a list of statements `[Statement]` because that's what the `suite` function in the Parser will give.

Here, you have some mismatching names, so it's not clear how you're trying to execute the statements. On line 54 you have the name `conseq`, but then on line 57 you use `body` and on line 58 you have `stmt`.

Sometimes the right names can help use make more sense of the code, so on line 54 try changing the name to `bodyStmts`, since that suggests more than one (i.e., a list) and those statements form the body of the loop.

You then need to evaluate all the statements in the list, sequentially, accumulating the changes to the store. You can do this recursively like the way it's done with `MultiStmt` in the HOPL.SIMPLE_STATEMENT language. Alternatively, someone else is going to be adding a helper function `resultOfStmts` that does this for a `[Statement]` - that hasn't been merged yet, but you could write your code to use that function since you know it's going to be in there. The type of that function is `resultOfStmts :: [Statement] -> Environment -> Store -> IO Store`.

Finally, once you have `st2` as a result of executing the `bodyStmts`, then you need to recursively run the loop by calling `resultOf` on the same `WhileStmt` again. See the implementation of `WhileStmt` in the HOPL.SIMPLE_STATEMENT language for how that should lookl

Tientuine (changes requested):

src/Adder/Interp.hs
DuncanMcLean: I updated it as you explained, the error is no longer a parse error but it is now telling me "Not in scope: data constructor 'BoolVal'." I'm not sure if this is an issue on my part or if it is something that isn't implemented into the language yet.

src/Adder/Interp.hs
DuncanMcLean: I made the changes as you suggested, it seems to compile but let me know if I am still missing anything.

DuncanMcLean:
I pushed the code for the "else" components of the while statement as well.

src/Adder/Interp.hs
```diff
@@ -51,6 +51,15 @@ resultOfProgram _ env st0 = undefined
 resultOf :: Statement -> Environment -> Store -> IO Store
 resultOf _ env st0 = undefined
 
+resultOf (WhileStmt test bodyStmts) p st0 = 
+  if q
+    then do 
+      st2 <- resultOf bodyStmts p st1
```
Tientuine, line 57: This should be `resultOfStmts` as we had discussed. I'm not sure how this could compile otherwise. I also don't see the `resultOfStmts` function anymore- I thought I remember seeing it before, but maybe I was mistaken?

src/Adder/Lang/Syntax.hs
```diff
@@ -26,7 +26,7 @@ data Statement
   | AugmentedAssignmentStmt Identifier AugOp Expression -- Bashir's Augmented Assignment constructor
   | BreakStmt
   | ContinueStmt
-  | WhileStmt Expression [Statement]
+  | WhileStmt Expression [Statement] [Statement]
```
Tientuine, line 29: Looks good, but with this change your implementation in Interp.hs is now broken because it doesn't handle the else statements.

src/Adder/Lang/Parser.hs
```diff
@@ -68,13 +68,14 @@ compoundStmt :: IParser Statement
...
     -- if_stmt ::=  "if" assignment_expression ":" suite   - If statement only
-    -- WhileStmt ::= "while" assignmentExpr ":" suite
+    -- WhileStmt ::= "while" assignmentExpr ":" suite ["else" : suite]
     [ IfStmt
         <$> (reserved "if" >> assignmentExpr)
         <*> (reservedOp ":" >> suite),
       WhileStmt
         <$> (reserved "while" >> assignmentExpr)
         <*> (reservedOp ":" >> suite)
+        <*> (reserved "else" >> suite)
```
Tientuine, line 78: This looks good, but note that all the parts have to work together (see my comment below for Syntax.hs).

src/Adder/Interp.hs
```diff
@@ -51,6 +51,15 @@ resultOfProgram _ env st0 = undefined
 resultOf :: Statement -> Environment -> Store -> IO Store
 resultOf _ env st0 = undefined
 
+resultOf (WhileStmt test bodyStmts) p st0 =
```
Tientuine, line 54: This will now have to be updated to handle the `else` statements since you added those to the syntax.

Tientuine (changes requested):
Good addition of the `else` part, but that also means that you need to update the interpreter implementation to support that new piece.

DuncanMcLean:
I made the changes you suggested, the only error I get an error for line 62 which says: "Not in scope: data constructor 'BoolVal'." I don't understand what causes this.

src/Adder/Interp.hs
```diff
@@ -51,6 +51,17 @@ resultOfProgram _ env st0 = undefined
...
+  if q
+    then do 
+      st2 <- resultOfStmts bodyStmts p st1
+      resultOf (WhileStmt test bodyStmts altern) p st2
+    else return st1
+    return st3    
+  where
+    (BoolVal q, st1) = valueOf test p st0
+    st3 = resultOfStmts altern p st1
+
```
Tientuine, line 64: Well done, this all looks correct, even the `else` part for now.

Tientuine (changes requested):
The rest of the message for the "Not in scope" error is: `Perhaps you want to add 'BoolVal' to the import list
    in the import of 'Adder.Lang.Syntax' (src\Adder\Interp.hs:23:1-72).` The fix is to do exactly what it says: add `BoolVal` to the import list. In fact, when you mouse over the red-underlined code, VSCode should suggest to you a few "Quick Fix" options, which you can just click on to have it add the necessary code. If you don't see that, maybe because the Haskell Language Server integration is not working, then you can just make the change manually.

The import line should read as either `import Adder.Lang.Syntax (Expression (..), Program (..), Statement (..), ExpVal(BoolVal))` or `import Adder.Lang.Syntax (Expression (..), Program (..), Statement (..), ExpVal(..))` depending on whether you want to import just the `BoolVal` constructor or all constructors for `ExpVal`.

Tientuine (approved):

PR #64, opened 2022-12-13
Mod and Float Division added to Interpreter

seanygberg:


src/Adder/Interp.hs
```diff
@@ -91,7 +91,17 @@ valueOf (BinaryExp op exp1 exp2) env st0 = valueOfBop op val1 val2
 
 valueOfBop :: BinaryOp -> ExpVal -> ExpVal -> ExpVal
 valueOfBop op val1 val2 = case op of
+  Mod -> case (val1, val2) of
+    (IntVal n1, IntVal n2) -> IntVal(n1 `mod` n2)
+    (IntVal n1, FloatVal n2) -> FloatVal(n1 `mod` n2)
+    (FloatVal n1, IntVal n2) -> FloatVal(n1 `mod` n2)
+    (FloatVal n1, FloatVal n2) -> FloatVal(n1 `mod` n2)
+    (StringVal n1, StringVal n2) -> undefined
+  Divide -> FloatVal(n1 `div` n2)
```
Tientuine, line 100: This will not work because the `n1` and `n2` defined below rely on a non-existent function, `expvalToNum`. See below for further explanation and suggestion for fixing this.

Tientuine (changes requested):
Great effort! You're very close to having this correctly implemented.

src/Adder/Lang/Syntax.hs
```diff
@@ -97,3 +97,10 @@ instance Show ExpVal where
   show (IntVal p) = show p
   show (FloatVal p) = show p
   show (StrVal p) = show p
+
+expvalToFloat :: ExpVal -> Float
+valueOf val -> case val of
```
Tientuine, line 102: You've declared the function name as `expvalToFloat` on line 101, but then on line 102 you are using the name `valueOf`. Line 102 should use the same function name as your type declaration on line 101. In fact, this should not have compiled... Did you do a `stack build` in the Terminal to make sure it compiles?

src/Adder/Interp.hs
```diff
@@ -91,7 +91,17 @@ valueOf (BinaryExp op exp1 exp2) env st0 = valueOfBop op val1 val2
 
 valueOfBop :: BinaryOp -> ExpVal -> ExpVal -> ExpVal
 valueOfBop op val1 val2 = case op of
+  Mod -> case (val1, val2) of
+    (IntVal n1, IntVal n2) -> IntVal(n1 `mod` n2)
+    (IntVal n1, FloatVal n2) -> FloatVal(n1 `mod` n2)
+    (FloatVal n1, IntVal n2) -> FloatVal(n1 `mod` n2)
+    (FloatVal n1, FloatVal n2) -> FloatVal(n1 `mod` n2)
+    (StrVal n1, StrVal n2) -> undefined
```
Tientuine, line 99: I recommend catching _all_ other possibilities here with the `_` pattern. This way, combinations like (IntVal,BoolVal), (BoolVal,StrVal), (FloatVal,BoolVal), etc. will all be handled in the final default case, so no one has to make separate cases for each of those combinations.

Tientuine (changes requested):
Nearly there! Just two teeny fixes left and then it's ready for approval and merge.

Tientuine (approved):
Good, done!

Issue #32, opened 2022-12-05
Float literals will never be parsed


Tientuine:
Integer and float literals can look similar. In fact, integer literals are effectively a subset of float literals.

For example, should the parser interpret `5` as an integer literal or a float literal? It makes sense that the parser should choose integer literal. What about `5.2`? Is this a float literal or an integer literal following by the `.` operator? Clearly, the parser should treat this as a float literal.

However, the current implementation of the `atom` parser tries to parse integer literals first. Since it will succeed in parsing an integer `5` when it encounters the token `5.2`, it will not parse it as a float. This is the wrong behavior. The `atom` parser should parse float literals first, only parsing integer literals if it didn't first parse a float.

IMPORTANT: It might seem like the solution is just a simple reordering of lines here, but there is one potential pitfall... When the re-ordered `atom` parser encounters `5`, it will try to parse it as a float literal, and it will only fail when it sees that there is no decimal point after the digit. But by that point, the parser has already consumed the digit `5`, so if we try to parse an integer literal we'll have nothing left of the token. The solution here is to `try` to parse the float literal - using the `try` function will effectively turn a parser into a look-ahead parser, meaning that if it fails then it will roll back and act like it didn't consume any characters of the token. To use the `try` function, just give it a regular, non-look-ahead parser - in other words, if `p` is a parser, then `try p` is a look-ahead version of that parser. (Note that here `p` may be a function or other expression that represents a parser.)

Stubeans:
![Screenshot (48) Zoomed](https://user-images.githubusercontent.com/79120702/208162443-943e2d84-22b0-4ac4-8ed6-0365b17a9b0a.png)
![Screenshot (47) error Zoomed](https://user-images.githubusercontent.com/79120702/208162448-1a212d52-7492-4f71-86e7-896659351040.png)

I'm curious why this doesn't work? Would I implement the try function here, or somewhere else? And then lets say I were to implement it here, how would I fix the precedence error between <|> and <$>? Do I convert to a <|> FloatLiteralExp >> float?

Stubeans:
![Screenshot (49) Zoomed](https://user-images.githubusercontent.com/79120702/208164664-1446a782-8054-42ed-9676-a60156fce7c6.png)

this is the only reference I see to p outside of the Parser, however I'm not entirely sure it refers to the parser. I tried putting a choice map here, as well as try methods but it doesn't seem to belong.

PR #65, opened 2022-12-14
Possible Fix - Update on Atom Implementation

Stubeans:
Sorry for the wait, I've got a lot of studying/assignments. I will hopefully implement new features as a part of the homework sometime tomorrow, ( or really today, now, Wednesday )

### DIFF
--- a/src/Adder/Interp.hs
+++ b/src/Adder/Interp.hs
@@ -24,7 +24,6 @@ import Adder.Lang.Syntax (Expression (..), Program (..), Statement (..))
 import Adder.Store (Store, deref, emptyStore, newref, setref)
 import Data.Either (fromRight)
 import Prelude hiding (exp)
-import GHC.Base (undefined)
 
 type Interpreter a = a -> Environment -> Store -> IO Store
 
@@ -73,15 +72,18 @@ type Answer = (ExpVal, Store)
 {- semantic reductions for expressions -}
 
 -- TODO Implement the semantics for each kind of Adder expression
-valueOf :: Expression -> Environment -> Store -> Answer
-valueOf _ env st0 = undefined
+
 -- Binary Operation
-valueOf (BinaryExp op exp1 exp2) env st0 = valueOfBop op val1 val2
+-- Don't forget about free store
+
+valueOf (BinaryExpr op exp1 exp2) env st0 = valueOfBop op val1 val2
   where
     (val1, st1) = valueOf exp1  env st0
     (val2, st2) = valueOf exp2 env st1
 
--- Don't forget about free store
+valueOf :: Expression -> Environment -> Store -> Answer
+valueOf _ env st0 = undefined
+
 
 --valueOF :: assignmentExpr ->  ??
 
@@ -89,7 +91,7 @@ valueOf (BinaryExp op exp1 exp2) env st0 = valueOfBop op val1 val2
 -- TODO Implement any helper functions needed to simplify the design of the
 -- interpreter (e.g., the applyProcedure helper function).
 
-valueOfBop :: BinaryOp -> ExpVal -> ExpVal -> ExpVal
+valueOfBop :: Answer -> ExpVal -> ExpVal -> ExpVal
 valueOfBop op val1 val2 = case op of
   _ -> error "unimplemented binary operation"
   --Code here


src/Adder/Interp.hs
```diff
@@ -71,26 +70,29 @@ type Answer = (ExpVal, Store)
...
 
 {- Auxiliary functions -}
 -- TODO Implement any helper functions needed to simplify the design of the
 -- interpreter (e.g., the applyProcedure helper function).
 
---valueOfBop :: BinaryOp -> ExpVal -> ExpVal -> ExpVal
---valueOfBop op val1 val2 = case op of
---  _ -> error "unimplemented binary operation"
---Code here
+valueOfBop :: Answer -> ExpVal -> ExpVal -> ExpVal
```
Tientuine, line 92: Not sure about the reason for this change, but it will not work. This needs to remain `BinaryOp`. The issue was that the symbol `BinaryOp` in not imported from Adder.Lang.Syntax. You can fix by either adding `BinaryOp(..)` to the import list above, or simply removing the import list (i.e., importing everything from the module).

src/Adder/Interp.hs
```diff
@@ -71,26 +70,29 @@ type Answer = (ExpVal, Store)
 {- semantic reductions for expressions -}
 
 -- TODO Implement the semantics for each kind of Adder expression
+
+-- Binary Operation
+-- Don't forget about free store
+
+valueOf (BinaryExpr op exp1 exp2) env st0 = valueOfBop op val1 val2
```
Tientuine, line 77: Calling `valueOfBop` gives you an `ExpVal`, but you'll also need the new free store. So, the right-hand side should be a pair containing both the result of `valueOfBop` and `st2`.

Tientuine (changes requested):
Two final revisions needed.

Tientuine (changes requested):
One issue fixed, but not the other. Rather than throwing stuff and seeing what sticks, let's take moment to break it down and understand the pieces.

On the one hand, you have your definition of `valueOfBop :: BinaryOp -> ExpVal -> ExpVal -> ExpVal`. This type means that the `valueOfBop` function takes in three arguments (a `BinaryOp` and two `ExpVal`s) and then returns an `ExpVal`.

Now, on line 77, you're trying to call `valueOfBop st2`... but `st2` has type `Store`. Note that in Haskell `f x` is the same as `f(x)`, calling function `f` and passing in argument `x`. So, you're not only passing in the wrong number of arguments (one instead of three), the one argument you're passing in is the wrong type (`Store` instead of `BinaryOp`).

What you had previously was `valueOfBop op val1 val2`. You're passing in three arguments, so that's good. But are they the right types? Well, `op` is a `BinaryOp` (since that what the first element of a `BinaryExpr` is). Then `val1` and `val2` are each the first elements of the pair returned by calling `valueOf`... since `valueOf` returns a pair of `(ExpVal, Store)`, the first element is an `ExpVal`, so we're good there. In other words, you were calling `valueOfBop` just fine already.

The remaining issue has nothing to do with how you're calling `valueOfBop`. The issue is that `valueOfBop` only returns an `ExpVal`, but you need a pair of `(ExpVal, Store)`. So, call `valueOfBop` the way you already were, but take what it gives you back and pair it with `st2`.

To clarify the syntax, imagine that we had a function `f` that takes an `Int` and returns a `Float`. Then imagine that I need a pair of `(Float, Bool)`. Then I can create one like this: `(f 5, True)`.

Tientuine:
@Stubeans We had an issue with the `for-merging` branch, and apparently it has corrupted your branch. I'm going to fix your branch, but then I will need you to delete your local branch and then pull your branch again.

After that, please fix one last error... you can't have your implementation for `valueOf` preceding its type declaration. That causes a problem. You should always `stack build` to make sure it compiles, then you would discover that there's an issue here.

Stubeans:
@Tientuine  I'm not sure what I just did to my repo, I thought i was working with a clean new version of my branch and then when I went to push my changes it asked to marge. I'm not sure If I merged it correctly however, could you take a look?

Stubeans:
I have also attempted to add elif and else to the current if statement to complete the homework. I commented it out, but does that look right?

Tientuine:
I'll take a look and see if I can figure out what happened. In the meantime, either hold off on other work or just make a new branch (e.g., `Smith-morework`) from `for-merging` and do any additional work on there.

Tientuine:
> I have also attempted to add elif and else to the current if statement to complete the homework. I commented it out, but does that look right?

Max is already working on those, but we had some branch issues tonight. Some of that work-in-progress was inadvertently merged before it was ready, which is why you may have seen snippets of that code get pulled into your branch earlier.

Stubeans:
@Tientuine I see that my Atom implementation was commented out of the Interp.hs file in for_merging. Should I stop working on the Atom component, and focus on a second implementation per the homework?

Which leads me to my second question, to avoid working on something that's already being done could you give me something to focus on trying to implement in this final day? I am very busy trying to complete another final project and I'm finding myself running out of time.

Tientuine:
> 

It's only commented out because it got merged prematurely but doesn't compile yet. Please make the last remaining for for the binary operations and then we can merge that in.

As for additional work, no one ever went back to fix the parse issue for integer and float literals, see Issue #32 . I expected that Cole might do it because he was the one who initially added support for literals, but if he's not taking it then that would be a good one to knock out.

Stubeans:
Screenshot (45) Zoomed

Now I know I have to work with the lines underlined in red, I'm just not exactly sure how to write the lines. i can see that the way the two lines 'talk' to each other is through the _ value that must represent an empty BinaryOp. Does this mean I have to write another line here with a different value than _, and set it equal to what? A BinaryExpr op exp1 exp2?

And then when I change the bottom portion, how do I implement another case for that? Am I writing the code that takes the resulting St1 and St2, putting them into an Answer?

Tientuine:
> Screenshot (45) Zoomed
> 
> Now I know I have to work with the lines underlined in red, I'm just not exactly sure how to write the lines. i can see that the way the two lines 'talk' to each other is through the _ value that must represent an empty BinaryOp. Does this mean I have to write another line here with a different value than _, and set it equal to what? A BinaryExpr op exp1 exp2?
> 
> And then when I change the bottom portion, how do I implement another case for that? Am I writing the code that takes the resulting St1 and St2, putting them into an Answer?

Here's the problem - once you define a function, you can't re-define it. The first line where you declare either the type of a function or define an implementation of the function (even just one case) effectively declares the function. Then, if you declare the type again, it's an error, think of it as a "split function definition." Consider the following example:

```haskell

-- This is okay
f :: Int -> Float -> Float
f 0 x = 0
f _ x = x

-- This is not okay
g 0 x = 0
g :: Int -> Float -> Float
g _ x = x
```

PR #66, opened 2022-12-14
Take 2 of if else stmt

MaxEnglish:
-Parser is yelling at me a lot and I'm not sure what's the matter with the code I have written there. 
-I remember you said that I'd only need to use resultOfStmts once per function, but I'm not sure how to do that so I just did what I was comfortable with. 
-I realize that you need an else stmt for every if-then-else in Haskell but for ifstmt and ifelif stmt I don't want an else. I believe that's why it's throwing an error for me. How can I circumvent this?

### DIFF
--- a/src/Adder/Interp.hs
+++ b/src/Adder/Interp.hs
@@ -20,7 +20,7 @@ import Adder.DataStructures (DenVal, Environment, ExpVal (..), Function (..))
 import Adder.Defs (Source)
 import Adder.Environment (Env (..))
 import Adder.Lang.Parser (ParseError, parseFile, parseInteractive)
-import Adder.Lang.Syntax (Expression (..), Program (..), Statement (..))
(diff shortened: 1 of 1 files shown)


Tientuine:
> -Parser is yelling at me a lot and I'm not sure what's the matter with the code I have written there. -I remember you said that I'd only need to use resultOfStmts once per function, but I'm not sure how to do that so I just did what I was comfortable with. -I realize that you need an else stmt for every if-then-else in Haskell but for ifstmt and ifelif stmt I don't want an else. I believe that's why it's throwing an error for me. How can I circumvent this?

I'll post comments on the files themselves shortly, but I think I may not have explained clearly enough about `resultOfStmts`. When I explained it early, it was in response to something you asked and I wanted to make sure that you didn't try to call the function twice (different from having it appear in the code twice). I did not mean that you would only write the function call once, but rather than it would only be called once. For example, consider
```haskell
    if test x then f y else f z
```
Here the function `f` is called only once, even though it appears twice in the code.

PR #68, opened 2022-12-14
English work

MaxEnglish:
-Parser is yelling at me a lot and I'm not sure what's the matter with the code I have written there. 
-I remember you said that I'd only need to use resultOfStmts once per function, but I'm not sure how to do that so I just did what I was comfortable with. 
-I realize that you need an else stmt for every if-then-else in Haskell but for ifstmt and ifelif stmt I don't want an else. I believe that's why it's throwing an error for me. How can I circumvent this?

### DIFF
--- a/src/Adder/Interp.hs
+++ b/src/Adder/Interp.hs
@@ -59,11 +59,35 @@ resultOf _ env st0 = undefined
 ---------------------------------------------
 -- env1 = env0
 {-
-resultOf (IfStmt test conseq) p st = if q then st2 else st3
+resultOf (IfStmt test conseq) p st = if q then st2
+  where
+    Answer (BoolVal q) st1 = valueOf test p st
+    st2 = resultOf conseq p st1
+-}
+
+{-
+resultOf (IfElseStmt test conseq) p st = if q then st2 else st3
   where
     Answer (BoolVal q) st1 = valueOf test p st
     st2 = resultOf conseq p st1
     st3 = resultOf conseq p st2
+-}
+
+resultOf (IfElifStmt test conseq) p st = if q then st2 then if r then st4
+  where
+    Answer (BoolVal q) st1 = valueOf test p st
+    st2 = resultOf conseq p st1
+    Answer (BoolVal r) st3 = valueOf test r st
+    st4 = resultOf conseq p st3
+-}
+
+resultOf (IfElifElseStmt test conseq) p st = if q then st2 then if r then st4 else st5
+  where
+    Answer (BoolVal q) st1 = valueOf test p st
+    st2 = resultOf conseq p st1
+    Answer (BoolVal r) st3 = valueOf test r st
+    st4 = resultOf conseq p st3
+    st5 = resultOf conseq p st4
 -}
 
 {- Evaluating a program yields an "answer" - a value and a resulting state. -}
--- a/src/Adder/Interp.hs
+++ b/src/Adder/Interp.hs
@@ -66,13 +66,13 @@ resultOf (IfStmt test conseq) p st = if q then st2
     st2 = resultOf conseq p st1
 -}
 
-{-
-resultOf (IfElseStmt test conseq) p st = if q then st2 else st3
+
+resultOf (IfElseStmt [exp1] [exp2] [exp3]) p st = if q then st2 else st3
   where
-    Answer (BoolVal q) st1 = valueOf test p st
-    st2 = resultOf conseq p st1
-    st3 = resultOf conseq p st2
--}
+    (BoolVal q, st1) = valueOf exp1 p st
+    st2 = resultOf exp2 p st2
+    st1 = resultOf exp3 p st3
+
 
 resultOf (IfElifStmt test conseq) p st = if q then st2 then if r then st4
   where
--- a/src/Adder/Interp.hs
+++ b/src/Adder/Interp.hs
@@ -25,6 +25,7 @@ import Adder.Store (Store, deref, emptyStore, newref, setref)
 import Data.Either (fromRight)
 import GHC.Base (undefined)
 import Prelude hiding (exp)
+import GHC.Float (stgWord32ToFloat)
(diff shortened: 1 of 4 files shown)


src/Adder/Interp.hs
```diff
@@ -57,13 +57,43 @@ resultOf (StmtList (stmt : stmts)) env st0 = do
...
-{-
-resultOf (IfStmt test conseq) p st = if q then st2 else st3
+resultOfStmts :: [Statement] -> Environment -> Store -> IO Store
+resultOfStmts [] p st1 = return st1
+resultOfStmts (stmt : stmts) p st1 = do
+  st2 <- resultOf stmt p st1
+  resultOfStmts stmts p st2
+
+
+resultOf (IfStmt exp1 stmt1) p st = if q then val2
```
Tientuine, line 67: Two issues here:

1. You're trying to use the name `val2` here and on line 70 below, but I don't see any `val2` declared. When the test `q` is true, what you need to return here is the state that results from executing the statements in the list `stmt1`, which down below you've named `s1`. So here, just return `s1` when `q` is true.
2. In Haskell, every if-expression must have an else branch. To implement an `IfStmt` without an `else` part, just make the else do nothing; i.e., when `q` is false, return the same state `st` that you started with.

One final comment on on coding style and consistency - we've generally used a numbering scheme to help understand the order of changes to the program state or free store. So, `st0` would represent the initial state, `st1` the next state, etc. To ensure consistency in the code base, let's try to stick to that naming convention.

src/Adder/Interp.hs
```diff
@@ -57,13 +57,43 @@ resultOf (StmtList (stmt : stmts)) env st0 = do
...
+  where
+    (BoolVal q, st) = valueOf exp1 p st
+    s1 = resultOfStmts stmt1 p val2
+
+
+
+resultOf (IfElseStmt exp1 trueStmts falseStmts) p st = if q then val2 else val3
```
Tientuine, line 74: Again, as above, you need to return the new states, `s1` or `s2`, instead of the undeclared `val2` and `val3`.

src/Adder/Interp.hs
```diff
@@ -57,13 +57,43 @@ resultOf (StmtList (stmt : stmts)) env st0 = do
...
+resultOfStmts :: [Statement] -> Environment -> Store -> IO Store
+resultOfStmts [] p st1 = return st1
+resultOfStmts (stmt : stmts) p st1 = do
+  st2 <- resultOf stmt p st1
+  resultOfStmts stmts p st2
+
+
+resultOf (IfStmt exp1 stmt1) p st = if q then val2
+  where
+    (BoolVal q, st) = valueOf exp1 p st
```
Tientuine, line 69: Since evaluating `exp1` might change the free store, you should have a name different from `st` on the left-hand side here. Maybe `st1`?

src/Adder/Interp.hs
```diff
@@ -57,13 +57,43 @@ resultOf (StmtList (stmt : stmts)) env st0 = do
...
+    s1 = resultOfStmts stmt1 p val2
```
Tientuine, line 70: Instead of `val2`, which has not been declared, the third argument on the right-hand side should be the state that resulted from evaluating `exp1`, which you should probably name `st1` for consistency. This would make the left-hand side the second new state, so a name like `st2` would be more appropriate.

src/Adder/Interp.hs
```diff
@@ -57,13 +57,43 @@ resultOf (StmtList (stmt : stmts)) env st0 = do
...
+    s2 = resultOfStmts p val4
+
+
+resultOf (IfElifElseStmt exp1 stmt1 exp2 stmt2 stmt3) p st = if q1 then val2 else if q2 then val4 else val5
+  where
+    (BoolVal q1, st1) = valueOf exp1 p st
+    s1 = resultOfStmts stmt1 p val2
+    (BoolVal q2, st2) = valueOf exp2 p st
+    s2 = resultOfStmts stmt2 p val4
+    s3 = resultOfStmts stmt3 p val5
```
Tientuine, line 95: Apply the same ideas from above to these cases to ensure that you're passing the correct free stores around, and that the naming convention is consistent.

src/Adder/Lang/Parser.hs
```diff
@@ -76,6 +76,23 @@ compoundStmt =
...
+        <*> (reservedOp ":" >> suite)
+        <*> (reserved "elif" >> assignmentExpr)
+        <*> (reservedOp ":" >> suite),
+      IfElifElseStmt
+        <$> (reserved "if" >> assignmentExpr)
+        <*> (reservedOp ":" >> suite)
+        <*> (reserved "elif" >> assignmentExpr)
+        <*> (reservedOp ":" >> suite)
+        <*> (reserved "else")
+        <*> (reservedOp ":" >> suite),
```
Tientuine, line 95: All this looks okay.

src/Adder/Lang/Syntax.hs
```diff
@@ -18,7 +18,10 @@ newtype Program
 
 -- TODO Build out the abstract syntax for Adder statements
 data Statement
-  = IfStmt Expression [Statement] -- If Statement contains Exprression and list of statements - at this time only one statement
+  = IfStmt Expression [Statement]
+  | IfElseStmt Expression [Statement] [Statement]
+  | IfElifStmt Expression [Statement] [Statement]
+  | IfElifElseStmt Expression [Statement] [Statement] [Statement]
```
Tientuine, line 24: This all looks fine for now.

Tientuine (changes requested):
Inching closer, but one more round of fixes is needed.

MaxEnglish:
Check "Prescribed Changes" commit for latest work

Tientuine (approved):
This all looks good, the logic is correct. Some of the variable names you used could be better and there is an extraneous import, but this is good work. Approved!

PR #69, opened 2022-12-15
Stengle work

jstengle15:


src/Adder/Interp.hs
```diff
@@ -87,11 +87,18 @@ valueOf _ env st0 = undefined
...
+
 --  _ -> error "unimplemented binary operation"
 --Code here
-
+valueOfBop :: BinaryOp -> ExpVal -> ExpVal -> ExpVal
+valueOfBop op val1 val2 = case op of
+  And -> BoolVal (x1 && x2)
+  Or -> BoolVal (m1 && m2)
+  where
+    x1 = Bool y1
```
Tientuine, line 98: Make sure to run `stack build` to compile the code and make sure there are no syntax or type errors. This line (and the following ones), are incorrect. One variation that will work is `BoolVal x1 = val1`, which means that when `val1` is a `BoolVal` you're going to use the name `x1` to refer to the `Bool` that's wrapped inside it.

Tientuine (changes requested):
Won't compile yet, but a few small changes should yield a working version. See below for a specific suggestion on one way to fix it.

src/Adder/Interp.hs
```diff
@@ -87,11 +87,18 @@ valueOf _ env st0 = undefined
...
-
+valueOfBop :: BinaryOp -> ExpVal -> ExpVal -> ExpVal
+valueOfBop op val1 val2 = case op of
+  And -> BoolVal (x1 && x2)
+  Or -> BoolVal (y1 && y2)
+  where
+    BoolVal x1 = val1
+    BoolVal x2 = val2
+    BoolVal y1 = val1
+    BoolVal y2 = val2
```
Tientuine, line 101: You really only need one set of these `BoolVal`s here, but I'll condense when I merge.

Tientuine (approved):

PR #70, opened 2022-12-16
Consiglio work - Added Interp support for add, concat, sub, mult, floor div, and powers

RichieConsiglio:


src/Adder/Interp.hs
```diff
@@ -63,3 +65,36 @@ valueOf _ env st0 = undefined
...
 -- TODO Implement any helper functions needed to simplify the design of the
 -- interpreter (e.g., the applyProcedure helper function).
+valueOfBop :: BinaryOp -> ExpVal -> ExpVal -> ExpVal
+valueOfBop op val1 val2 = case op of
+  Plus -> case (val1, val2) of
+    (IntVal n1, IntVal n2) -> IntVal(n1 'plus` n2)
+    (IntVal n1, FloatVal n2) -> FloatVal(n1 `plus` n2)
+    (FloatVal n1, IntVal n2) -> FloatVal(n1 `plus` n2)
+    (FloatVal n1, FloatVal n2) -> FloatVal(n1 `plus` n2)
+    (StrVal n1, StrVal n2) -> StrVal(n1 `plus` n2)
```
Tientuine, line 75: Haskell has no `plus` function for adding things, just the `+` operator, so you should just use that. Also, for joining strings you cannot use `+`, which is for numbers only. Instead you can use `++`.

One more note on style/naming... Names like `n1` and `n2` make sense for numbers (like whats in the `IntVal`s and `FloatVal`s) but they are strange names for strings, which is whats in an `StrVal`. For the `StrVal`s, maybe names like `s1` and `s2` would make more sense.

src/Adder/Interp.hs
```diff
@@ -63,3 +65,36 @@ valueOf _ env st0 = undefined
...
+    _ -> undefined
+  Minus -> case (val1, val2) of
+    (IntVal n1, IntVal n2) -> IntVal(n1 `minus` n2)
+    (IntVal n1, FloatVal n2) -> FloatVal(n1 `minus` n2)
+    (FloatVal n1, IntVal n2) -> FloatVal(n1 `minus` n2)
+    (FloatVal n1, FloatVal n2) -> FloatVal(n1 `minus` n2)
```
Tientuine, line 81: Similar to above, Haskell does not have a `minus` function for doing subtraction. Instead, just use the `-` operator as usual.

src/Adder/Interp.hs
```diff
@@ -63,3 +65,36 @@ valueOf _ env st0 = undefined
...
+    _ -> undefined
+  Times -> case (val1, val2) of
+    (IntVal n1, IntVal n2) -> IntVal(n1 `times` n2)
+    (IntVal n1, FloatVal n2) -> FloatVal(n1 `times` n2)
+    (FloatVal n1, IntVal n2) -> FloatVal(n1 `times` n2)
+    (FloatVal n1, FloatVal n2) -> FloatVal(n1 `times` n2)
```
Tientuine, line 87: Likewise, no `times` function in Haskell. Just use the `*` operator.

src/Adder/Interp.hs
```diff
@@ -63,3 +65,36 @@ valueOf _ env st0 = undefined
...
+    _ -> undefined
+  FloorDiv -> case (val1, val2) of
+    (IntVal n1, IntVal n2) -> IntVal(n1 `floordiv` n2)
+    (IntVal n1, FloatVal n2) -> FloatVal(n1 `floordiv` n2)
+    (FloatVal n1, IntVal n2) -> FloatVal(n1 `floordiv` n2)
+    (FloatVal n1, FloatVal n2) -> FloatVal(n1 `floordiv` n2)
```
Tientuine, line 93: Again, no `floordiv` function in Haskell.

For integers, Haskell does provide a `div` function that you can use. But that will only work if both operands are integers.

In the other cases, you should do regular division of floats with `/` and then use the `floor` function on the result. Of course, this only works with two floats, so if one operand is an integer then you'll need to convert it to a float. It's easy to look up how to do this in Haskell, but to save you time, the name of the function you'll need is `fromInteger`.

src/Adder/Interp.hs
```diff
@@ -63,3 +65,36 @@ valueOf _ env st0 = undefined
...
+    _ -> undefined
+  Power -> case (val1, val2) of
+    (IntVal n1, IntVal n2) -> IntVal(n1 `power` n2)
+    (IntVal n1, FloatVal n2) -> FloatVal(n1 `power` n2)
+    (FloatVal n1, IntVal n2) -> FloatVal(n1 `power` n2)
+    (FloatVal n1, FloatVal n2) -> FloatVal(n1 `power` n2)
```
Tientuine, line 99: No `power` function, unfortunately. But it does have a `**` operator. However, this operator will only work for floats. So, if you have two integers, you'll need to convert them to floats (see above), do the power operation, and then convert the result back to an integer (you can use the `floor` function for this).

src/Adder/Lang/Lexer.hs
```diff
@@ -40,7 +40,7 @@ adderLexSpec =
       Tok.opStart = Tok.opLetter adderLexSpec,
       Tok.opLetter = oneOf "+-*/%@<>&|^~:=!.",
       Tok.reservedOpNames =
-        []
+        ["Add","Concat"]
```
Tientuine, line 43: These are not correct. You should have `+`, `-`, `*`, `//`, and `**` here instead.

Tientuine (changes requested):

src/Adder/Interp.hs
```diff
@@ -68,29 +68,29 @@ valueOf _ env st0 = undefined
...
     _ -> undefined
   FloorDiv -> case (val1, val2) of
-    (IntVal n1, IntVal n2) -> IntVal(n1 `floordiv` n2)
-    (IntVal n1, FloatVal n2) -> FloatVal(n1 `floordiv` n2)
-    (FloatVal n1, IntVal n2) -> FloatVal(n1 `floordiv` n2)
-    (FloatVal n1, FloatVal n2) -> FloatVal(n1 `floordiv` n2)
+    (IntVal n1, IntVal n2) -> IntVal(floor(fromIntegral n1 `/` fromIntegral n2))
+    (IntVal n1, FloatVal n2) -> FloatVal(floor(fromIntegral n1 `/` fromIntegral n2)))
+    (FloatVal n1, IntVal n2) -> FloatVal(floor(fromIntegral n1 `/` fromIntegral n2)))
+    (FloatVal n1, FloatVal n2) -> FloatVal(floor(n1 `/` n2))
```
Tientuine: The backticks are only used if you're calling a named function in the infix style (between its operands). For regular symbolic operations, just put the operator without the backticks, as in `n1 + n2`.

src/Adder/Lang/Lexer.hs
```diff
@@ -38,9 +38,9 @@ adderLexSpec =
       -- TODO Define the reserved operator symbols for the Adder language
       -- See https://docs.python.org/3/reference/lexical_analysis.html#operators
       Tok.opStart = Tok.opLetter adderLexSpec,
-      Tok.opLetter = oneOf "+-*/%@<>&|^~:=!.",
+      Tok.opLetter = oneOf "+,-,*,//,**",
```
Tientuine, line 41: This line actually did not need any changes. This should be changed back to what it was.

src/Adder/Lang/Lexer.hs
```diff
@@ -38,9 +38,9 @@ adderLexSpec =
...
       Tok.reservedOpNames =
-        ["Add","Concat"]
+        ["Plus","Minus","Times","FloorDiv","Power"]
```
Tientuine, line 43: The point of this code in the Lexer is to specify what the operators look like in the Adder language itself. The abstract syntax names from the Syntax.hs file are just for internal use in the interpreter and should not appear in the Lexer. Here, just put the operator symbols themselves, as in `["+", "-", "*", "//", "**"]`.

Tientuine (changes requested):

src/Adder/Lang/Parser.hs
```diff
@@ -99,3 +101,15 @@ atom :: IParser Expression
...
+-- EBNF form: plus := +
+--            concat := +
+
+
+operator :: IParser BinaryOp
+operator = 
+  (choice . map try)
+  [ Plus <$ reservedOp "+",
+    Concat <$ reservedOp "+"
+  ]
```
Tientuine, line 115: I don't think this is needed, as I don't see anywhere else in the code where you use this `operator` function.

src/Adder/Interp.hs
```diff
@@ -63,3 +65,36 @@ valueOf _ env st0 = undefined
...
+    _ -> undefined
+  FloorDiv -> case (val1, val2) of
+    (IntVal n1, IntVal n2) -> IntVal(floor(fromIntegral n1 // fromIntegral n2))
+    (IntVal n1, FloatVal n2) -> IntVal(floor(fromIntegral n1 // n2))
+    (FloatVal n1, IntVal n2) -> IntVal(floor(n1 // fromIntegral n2))
+    (FloatVal n1, FloatVal n2) -> IntVal(floor(n1 // n2))
+    _ -> undefined
+  Power -> case (val1, val2) of
+    (IntVal n1, IntVal n2) -> IntVal(floor(fromIntegral n1 ** fromIntegral n2))
+    (IntVal n1, FloatVal n2) -> FloatVal(floor(fromIntegral n1 ** n2))
```
Tientuine: For this case, the result should be a float, so just don't use `floor` to convert back to `IntVal`.

Did you try to `stack build` in the Terminal? I don't think this code would compile.

Tientuine (changes requested):
I think just one more small change is needed. (And maybe delete an unnecessary function definition in the Parser.hs file.)

Tientuine (approved):

PR #71, opened 2022-12-16
Smith morework - Atom Implementation Continued

Stubeans:
About to push changes...


Stubeans:
So I moved the valueOfBop above where it's being called, and added a case. However, I realize that the case may be accidentally recursive. Do I have the right idea?

I also consolidated the valueOf statement, and moved the valueOf( BinaryExpr...)... above the undefined case.

Stubeans:
I fixed a build issue but I'm not sure If i did it correctly. I'm really not sure at all about what valueOfBop does exactly, I assumed I had to call BinaryExpr somehow to reach that code.

Stubeans:
And I believe I got the try in the right place for the literals fix

src/Adder/Lang/Parser.hs
```diff
@@ -164,12 +164,13 @@ assignmentExpr = expression -- For now, assignment expression only needs to be a
 -- See https://docs.python.org/3/reference/expressions.html#grammar-token-python-grammar-atom
 atom :: IParser Expression
 atom =
-  AtomExp . IdAtom <$> identifier
+  AtomExp . try IdAtom <$> identifier
```
Tientuine, line 167: Actually, the `identifier` parser isn't the problem, so you can take that out. It's the parser for float literals that's the problem, that's where you need to add the `try`. However, the `try` function takes a parser, so you'll need to wrap the float literal parser in parentheses to to make sure Haskell does the order of operations correctly... e.g., `try (FloatLiteralExp <$> float)`.

src/Adder/Interp.hs
```diff
@@ -71,26 +71,23 @@ type Answer = (ExpVal, Store)
 {- semantic reductions for expressions -}
 
 -- TODO Implement the semantics for each kind of Adder expression
+valueOfBop :: BinaryOp -> ExpVal -> ExpVal -> (ExpVal, Store)
+valueOfBop op val1 val2 = case op of
+  _ -> (val1, emptyStore)
+  _ -> error "unimplemented binary operation"
+  --Code here
+
```
Tientuine, line 79: Although it won't cause a problem here, I specifically had a comment below for where we should put auxiliary/helper functions. So, let's keep them down there. Please put this back down below and put the `{- Auxiliary functions -}` comment back in also. Thanks!

src/Adder/Interp.hs
```diff
@@ -71,26 +71,23 @@ type Answer = (ExpVal, Store)
 {- semantic reductions for expressions -}
 
 -- TODO Implement the semantics for each kind of Adder expression
+valueOfBop :: BinaryOp -> ExpVal -> ExpVal -> (ExpVal, Store)
+valueOfBop op val1 val2 = case op of
+  _ -> (val1, emptyStore)
```
Tientuine, line 76: There's no need for this case... your error case is enough here.

src/Adder/Interp.hs
```diff
@@ -71,26 +71,23 @@ type Answer = (ExpVal, Store)
...
+  _ -> error "unimplemented binary operation"
+  --Code here
+
 valueOf :: Expression -> Environment -> Store -> Answer
+valueOf (BinaryExpr op exp1 exp2) env st0 = valueOfBop op val1 val2
```
Tientuine, line 81: You still need to return the free store on the right-hand side along with the result of `valueOfBop`. That's still missing. Just take your whole `valueOfBop` function call with all its arguments and make that the first element of a pair, where the second element of the pair is the most recent free store, which in your case is `st2` since that's the one you ended up with after evaluating `exp2`.

Tientuine (changes requested):

src/Adder/Interp.hs
Stubeans: I'm about to push the change, however I don't quite understand why it's on this line. I thought that the right side of the = was a way of saying to pass the values on the left side into the values on the right, not that we're saying that the answer to the BinaryExpr is the right side. Did I explain that right? So we're basically saying that, given these expressions, where < the logic >, the answer is = to this?

Also, as a side note. When I changed the right side to an answer pair, I had to change the valueOfBop to return an ExpVal since that's what the answer was expecting. I hope that's correct.

Tientuine: If I'm understanding you correctly, then I think I understand where the confusion is.

When we define a function in Haskell, just like any other language the right-hand side describes the result.
```haskell
-- This function says that if we call `f` on some Double value that we'll name `x`,
-- then it will return another Double value that is equivalent to that `x` squared plus 1.
f :: Double -> Double
f x = x**2 + 1
```

I think maybe the confusion arises when we use the `where` clause to define other names. For example,
```haskell
-- This function says that if we call `g` on some Double value that we'll name `x`,
-- then it will return another Double value that is equivalent to that `x` squared plus 1.
-- But this function does so by naming the `x` squared part as `w` first, and then returning `w` plus 1.
g :: Double -> Double
g x = w + 1
  where
    w = x**2
```

In any case, the type name that appears last (i.e., farthest to the right) in our function type is the return type. So whatever is on the right-hand side of a function definition has to match that type.

```haskell
-- This type of this function says that it returns a pair of integer and double, so
-- the right-hand side must be a pair of integer and double. In this particular
-- case, the function `h` takes `x` and `y` and returns a pair containing `x` plus 1
-- as the first element of the pair and `y` divided by 2 as the second.
h :: Integer -> Double -> (Integer, Double)
h x y = (x + 1, y / 2)
```

src/Adder/Interp.hs
```diff
@@ -71,31 +71,32 @@ type Answer = (ExpVal, Store)
 {- semantic reductions for expressions -}
 
 -- TODO Implement the semantics for each kind of Adder expression
+valueOfBop :: BinaryOp -> ExpVal -> ExpVal -> ExpVal
+valueOfBop op val1 val2 = case op of
+  _ -> error "unimplemented binary operation"
+  --Code here
```
Tientuine, line 77: I would still like to see this function definition placed down below under the `{- Auxiliary functions -}` comment like you had it originally.

src/Adder/Interp.hs
```diff
@@ -71,31 +71,32 @@ type Answer = (ExpVal, Store)
...
+
+-- Binary Operation
+-- Don't forget about free store
+
 valueOf :: Expression -> Environment -> Store -> Answer
+valueOf (BinaryExpr op exp1 exp2) env st0 = (valueOfBop op val1 val2, st2)
```
Tientuine, line 83: This is correct now, because `valueOfBop` returns an `ExpVal` and `st2` is a free `Store`, so the right-hand side now produces a pair of those two things.

src/Adder/Lang/Parser.hs
```diff
@@ -164,12 +164,13 @@ assignmentExpr = expression -- For now, assignment expression only needs to be a
...
+    <|> FloatLiteralExp <$> float
```
Tientuine, line 168: Just need to fix this part by removing the `try` from the `IdAtom` rule and instead moving it to the float literal rule.

src/Adder/Lang/Syntax.hs
```diff
@@ -92,8 +92,9 @@ data Atom
   --Will need Enclosure
 
 -- TODO Implement "to-string" functionality for new Adder expressed values
-instance Show ExpVal where
+instance Show ExpVal where 
+  show (FloatVal p) = show p
```
Tientuine, line 96: This didn't need to be moved, but that's okay, we can leave it here.

Tientuine (changes requested):

Stubeans:
Should be all good now. I was able to build and run.

Tientuine (approved):

PR #72, opened 2022-12-16
Gjidoda_HW-4_Adder-pt2

Gjidodaj02:
First, I'm not entirely sure if I represented n1 and n2 as an int, spring or float correctly. 
Second, I understand the IsNot expression using "/=" to compare the n1 and n2. Which I also used for NotEqual since it said to in the issues, however, why wouldn't it just use "!=" ?

### DIFF
--- a/src/Adder/Interp.hs
+++ b/src/Adder/Interp.hs
@@ -59,7 +59,30 @@ type Answer = (ExpVal, Store)
 -- TODO Implement the semantics for each kind of Adder expression
 valueOf :: Expression -> Environment -> Store -> Answer
 valueOf _ env st0 = undefined
-
+valueOf (LessExp exp₁ exp₂) env st0 = BoolVal (n₁ < n₂)
+  where
+    n₁ = valueOf exp₁ env st0  
+    n₂ = valueOf exp₂ env st0
+valueOf (LessEqualExp exp₁ exp₂) env st0 = BoolVal (n₁ <= n₂)
+  where
+    n₁ = valueOf exp₁ env st0  
+    n₂ = valueOf exp₂ env st0
+valueOf (EqualExp exp₁ exp₂) env st0 = BoolVal (n₁ == n₂)
+  where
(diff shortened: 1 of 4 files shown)


Trent325:
i think != is \= in haskell

test.add
```diff
@@ -14,3 +14,19 @@ if computation == "area":
...
+print(x <= y)
+print(x <= z)
+print(x == y)
+print(x == z)
+print(x != y)
+print(x != z)
+print(x is y)
+print(x is z)
+print(x is not y)
+print(x is not z)
```
Tientuine, line 32: Put this in your own test file! Name it something that indicates what it's testing, like `test-inequalities.add` or something like that.

src/Adder/Interp.hs
```diff
@@ -59,7 +59,30 @@ type Answer = (ExpVal, Store)
...
+    n₁ = valueOf exp₁ env st0  
+    n₂ = valueOf exp₂ env st0
+valueOf (IsExp exp₁ exp₂) env st0 = BoolVal (n₁ == n₂)
+  where
+    n₁ = valueOf exp₁ env st0  
+    n₂ = valueOf exp₂ env st0
+valueOf (IsNotExp exp₁ exp₂) env st0 = BoolVal (n₁ /= n₂)
+  where
+    n₁ = valueOf exp₁ env st0  
+    n₂ = valueOf exp₂ env st0
```
Tientuine, line 85: This approach could work, except the Parser is not written to build separate kinds of AST nodes for the various kinds of comparisons. Instead, the Parser treats them all as `BinaryExpr`, distinguishing each one by its `BinaryOp`.

See the implementations of `Mod`, `Divide`, etc. in the latest **for-merging** branch to see how this should work.

Tientuine (changes requested):
The approach your taking could be a valid one if we were parsing these kinds of expressions separately, but actually our Adder parser is built different. See below for more.

PR #73, opened 2022-12-16
Comeau work

macc125:
A theoretical definition of Unary operators with cases of `Negative`, `Positive` and `Not`.


src/Adder/Interp.hs
```diff
@@ -73,7 +73,10 @@ type Answer = (ExpVal, Store)
 
 -- TODO Implement the semantics for each kind of Adder expression
 valueOf :: Expression -> Environment -> Store -> Answer
-valueOf _ env st0 = undefined
+valueOf (UnaryExpr op exp) env st0 = do   -- Helper function to check each unary op, and return
+  let (expVal, st1) = valueOf exp env st0 -- correct values
+  return (valueOfUop op expVal, st1)
+
```
Tientuine, line 79: You don't need the `do`-notation here. You can get rid of `do` and `return`, and instead put the keyword `in` before your final pair `(valueOfUop op expVal, st1)`.

src/Adder/Lang/Parser.hs
```diff
@@ -22,6 +22,9 @@ import Text.Parsec hiding (parse, string)
 import Text.Parsec.Expr
 import Text.Parsec.Indent
 
+-- Data type for boolean expressions
+data BoolExpr = TrueExp | FalseExp
+
```
Tientuine, line 27: I don't see this used anywhere else, so I don't think you need it. This can be removed.

src/Adder/Lang/Syntax.hs
```diff
@@ -44,6 +44,10 @@ data AugOp
 data UnaryOp
   = Not
   | Negative
+  | Positive
+  | BoolVal
+  | IntVal
+  |
```
Tientuine, line 50: These are not needed and should be removed. The `BoolVal` and `IntVal` constructors are already defined under the `ExpVal` data type.

src/Adder/Interp.hs
```diff
@@ -84,8 +87,30 @@ valueOf _ env st0 = undefined
...
+valueOfUop op exp st0 = case op of
+  Negative -> case valueOf exp st0 of
+    IntVal n -> IntVal (negate n)
+    BoolVal b -> error "Invalid operand for unary operator '-'"
+  Positive -> case valueOf exp st0 of
+    IntVal n -> IntVal n
+    BoolVal b -> error "Invalid operand for unary operator '+'"
+  Not -> case valueOf exp st0 of
+    IntVal n -> error "Invalid operand for unary operator 'not'"
+    BoolVal b -> BoolVal (not b)
```
Tientuine, line 116: The type of this function is incorrect. You are already computing the `valueOf` the expression up above on line 77, and then you're passing in the resulting `ExpVal` into this function. So this function needs only the `UnaryOp` and the `ExpVal`, so its type should be `valueOfUop :: UnaryOp -> ExpVal -> ExpVal`.

As a result, `exp` should probably be renamed something like `val` (since it represents an `ExpVal`) and you can remove `st0` entirely. Then, you don't need to recompute the `valueOf` anything here, instead you just need to take the `val` and do the appropriate operation with it.

Tientuine (changes requested):
Always try to `stack build` to make sure your code compiles before requesting review. If it doesn't compile, then it's not going to be ready for approval and merge. If you just want feedback, it's possible to make your PR a "draft pull request" by clicking the little downward arrow on the green button when creating the PR.
image

